PR #23, opened 2026-04-30
Add OpenCode compatibility

tksuns12:
## Summary
- add OpenCode AGENTS.md docs, native plugin hook generation, and OpenCode skills/agents output
- add model-resolution layer that maps hyperresearch opus/sonnet/haiku roles to active OpenCode provider/model IDs, preferring flat-rate equivalents
- add install platform selection so users can install Claude Code, OpenCode, or both via interactive checklist or --platform

## Verification
- .venv/bin/ruff check src tests
- .venv/bin/python -m pytest -q

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,13 @@ draco-bench/
 # in src/hyperresearch/skills/ (skills) and src/hyperresearch/core/hooks.py
 # (agent templates).
 .claude/
+.opencode/
 CLAUDE.md
 AGENTS.md
 GEMINI.md
+
+# Keep the repository-level OpenCode guide tracked.
+!/AGENTS.md
 
 # Per-vault research output
 research/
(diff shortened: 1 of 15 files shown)


jordan-gibbs:
Thanks. the architecture here is solid (Claude-compat skill discovery instead of duplicating the skill tree, real Task/Skill primitives on OpenCode, bun-executed plugin test) and I'd merge it once one thing is fixed: the opencode_permission_block translation in core/hooks.py emits allow for read/edit/bash/task but never deny for everything else, which silently breaks the patcher's and polish-auditor's tools: [Read, Edit] tool-lock, the load-bearing "patch, never regenerate" invariant on Claude's side. Please make the permission block default-deny (explicit write: deny, bash: deny, webfetch: deny for any agent whose Claude-side tools: line is Read, Edit) so  the invariant carries to OpenCode rather than degrading to a prompt-only nudge.

PR #26, opened 2026-05-08
Add Codex project adapter

MKrbm:
## Summary

Adds first-class Codex project installation. New `hyperresearch install --codex`
generates Codex-facing `AGENTS.md`, the entry skill + 16 step skills under
`.agents/skills/`, custom agents under `.codex/agents/*.toml`, and project-local
hooks under `.codex/`. Backend remains the existing `hyperresearch ... --json`
CLI; the vault model is unchanged.

## Compatibility

Claude workflow is the parent source of truth and is not regressed.

- Bundled Claude skills and subagent definitions are byte-identical to `main`.
- Default `hyperresearch install`, `install --global`, `install --steps-only`
  remain Claude paths.
- Codex artifacts are generated from the bundled Claude definitions plus an
  adapter preamble and mechanical translations.

| Shared-code change | Claude impact |
|---|---|
| `cli/install.py` `--codex` flag | None — gated by early return |
| `Vault.init(inject_docs=...)` | None — defaults to existing behavior |
| `agent_docs.py` Codex injector | None — separate marker pair |
| `search/fts.py` empty-query branch | None — non-empty path unchanged |
| `core/codex.py`, `codex_model_map.yaml` | None — new files |
| `core/sync.py` `research/temp/` filter | **Affects Claude too** — intended fix; details below |

## Verification

Passed on 2026-05-08:

- `pytest tests/ -q`, `ruff check src tests`
- `uv build` → wheel/sdist package-data inspection
- 0.9.0 wheel install in a fresh uv Python 3.13 venv
- `hyperresearch --version` and `hpr --version` from the installed wheel
- Packaged `hyperresearch install --codex` smoke
- Generated Codex hook script direct SessionStart smoke
- Live Codex run (`codex-cli 0.129.0`, `codex exec --skip-git-repo-check
  --sandbox workspace-write --json`) against the packaged install:

```text
AGENTS_FILE=yes      ENTRY_SKILL_FILE=yes  HOOKS_FILE=yes
CONFIG_FILE=yes      FETCHER_AGENT_FILE=yes
SKILL_COUNT=17       AGENT_COUNT=14        STATUS_OK=true
NOTES_TOTAL=0
```

### What changed

- `hyperresearch install --codex` (writes `AGENTS.md`, `.agents/skills/`,
  `.codex/agents/*.toml`, `.codex/config.toml`, `.codex/hooks.json`,
  `.codex/hooks/hyperresearch_pre_tool_use.py`).
- Data-driven Claude-label → Codex-model map at
  `src/hyperresearch/codex_model_map.yaml`.
- Installer output explains Codex project trust requirements.
- README install docs for both Claude Code and Codex.
- 0.9.0 changelog and version bump.

### Backend boundary

No MCP rewrite, no storage rewrite. Codex calls `hyperresearch ... --json` for
every backend operation, same as Claude. The Codex port is a parity adapter,
not a separate orchestration backend.

### Shared backend hardening

Two changes apply to both Claude and Codex; they surfaced during Codex parity
dry runs and clean up Claude behavior:

- **Empty filtered searches**: `hyperresearch search "" --tag <tag> --json`
  now lists structured results instead of passing `""` into SQLite FTS5
  `MATCH`.
- **`research/temp/` two-category sync**: `temp/` holds both real notes
  (broken-link stubs from `repair`, with frontmatter — must sync so wiki-links
  resolve) and workflow scratch (`draft-a.md`, `synthesis-plan.md`, etc., no
  frontmatter, addressed by direct path — must NOT sync). Pre-fix everything
  synced, so workflow scratch leaked into `note list` / `search` and could be
  rewritten by repair. New `_is_temp_workflow_artifact` filter
  (`core/sync.py`) skips workflow scratch by name list, prefix
  (`interim-report-`, `source-analysis-`), or absent frontmatter; frontmatter
  stubs continue to sync. Tests cover both directions and assert the denylist
  matches the workflow contract so it cannot silently drift.

### Codex runtime notes

- Project-local hooks load only after Codex trusts the project. The generated
  `AGENTS.md` and install output explain this and include a trusted-project
  TOML snippet.
- Generated Codex config uses the current stable feature flag:
  `[features]\nhooks = true`.
- Tool-lock parity is enforced via generated instructions, sandbox settings,
  hooks, and lint checks — not exact Claude allowlist equivalence.

### Packaging

The 0.9.0 sdist excludes local Codex-porting scratch files. The wheel
includes the Codex model map and bundled skill source files needed by
`install --codex`.

### Known limits

- Live external source quality and authenticated Crawl4AI profile behavior
  still depend on the existing fetch backend.
- Codex hooks are guardrails / context injection, not hard security
  boundaries.
- Non-interactive `codex exec` smoke commands need `--sandbox workspace-write`
  because SQLite may create WAL/cache files even for read-looking commands.

## Follow-ups (out of scope for 0.9.0)

1. **Split workflow scratch from sideline notes.** Move workflow scratch to a
   new `research/workspace/` directory and revert `temp/` to "sideline notes
   only". Removes the hard-coded denylist and the namespace-collision edge
   case where a wiki-link target ID happens to match a workflow-artifact name.
   Non-trivial: 16 skill files plus the Codex adapter reference
   `research/temp/<artifact>.md` paths directly and need a per-call audit.
2. **Single source of truth for the PreToolUse hook script.** Claude
   (`HOOK_SCRIPT_TEMPLATE`, JS) and Codex (`CODEX_HOOK_SCRIPT_TEMPLATE`,
   Python) currently have two independent copies of the same detection logic
   with no parity test. Extract shared rules into a YAML table or expose the
   classifier through `hpr` so both scripts shell out to it.
3. **Centralize the Codex agent registry.** Adding a subagent today requires
   editing `core/hooks.py` (twice) plus the hard-coded list in
   `core/codex.py:436` `_codex_agent_sources()`. Forgetting the third edit
   silently drops the agent on Codex installs. Replace with iteration over a
   single registry next to the agent constants.
4. **User-overridable Codex model map.** `codex_model_map.yaml` ships baked
   into the wheel with no CLI flag, env var, or local override. Let a project
   local `.codex/model_map.yaml` (if present) override the bundled map so
   users can react to Codex model renames without waiting for a new release.


jordan-gibbs:
Thanks for the well-scoped PR and the genuine effort here — the shared-backend fixes (empty FTS query handling, research/temp/ workflow denylist in sync) are real wins and I'll cherry-pick those separately. But I'm going to decline the Codex adapter itself, because the architectural mismatch is load-bearing rather than cosmetic. Your own CODEX_AGENT_PREAMBLE admits it: "if Codex cannot enforce the same tool surface exactly, treat the original tools line as a behavioral restriction" — that downgrades the patcher's and polish-auditor's tools: [Read, Edit] from a physical allowlist (Claude's harness blocks Write calls) to a prompt the model may ignore under context pressure. The whole "patch, never regenerate" guarantee that hyperresearch sells on the README rests on that being enforcement, not vibes. Beyond that, the V8 architecture's reliability gain came specifically from on-demand fresh procedure loading via the Skill tool (which Codex has no equivalent of), parallel subagent dispatch with per-agent model routing (the PR's .codex/agents/*.toml don't demonstrate Codex actually spawns concurrent custom-agents with distinct models), and your "verification" output — SKILL_COUNT=17 AGENT_COUNT=14 STATUS_OK=true NOTES_TOTAL=0 — only confirms files exist; there's no end-to-end Codex run producing a final report comparable to a Claude run. Shipping this as advertised would put the README's "leads the DeepResearch-Bench leaderboard" claim on a degraded pipeline that doesn't enforce the invariants the leaderboard rewards. I'd happily revisit if Codex grows true per-spawn tool-locking and you can show a side-by-side run that matches Claude's output quality on a representative query.

PR #36, opened 2026-06-22
fix(web): engage patchright stealth by passing UndetectedAdapter to the crawler

### LINKED ISSUE #35: crawl4ai provider never engages stealth — AsyncWebCrawler built with no crawler_strategy, so plain Playwright is always used instead of patchright
### Summary

`Crawl4AIProvider` constructs `AsyncWebCrawler(config=self._browser_config)` with no `crawler_strategy`. crawl4ai's `AsyncPlaywrightCrawlerStrategy` defaults its `browser_adapter` to `PlaywrightAdapter()` — i.e. plain Playwright. patchright's stealth engine only engages when an `UndetectedAdapter` is passed explicitly. So the provider's anti-bot behavior is effectively just `simulate_user=True` + the smart-wait JS; the patched-Chromium stealth driver is never used.

### Where

`src/hyperresearch/web/crawl4ai_provider.py`, two call sites:
- `Crawl4AIProvider._fetch_async` — `async with AsyncWebCrawler(config=self._browser_config) as crawler:`
- `Crawl4AIProvider._fetch_many_async` — `async with AsyncWebCrawler(config=self._browser_config) as crawler:`

### Steps to reproduce

Compare the runtime driver module with the default wiring vs. an explicit `UndetectedAdapter`:

```python
from crawl4ai import AsyncWebCrawler, BrowserConfig
from crawl4ai.async_crawler_strategy import AsyncPlaywrightCrawlerStrategy
from crawl4ai.browser_adapter import UndetectedAdapter, PlaywrightAdapter
import asyncio

async def driver(adapter):
    bc = BrowserConfig(headless=True)
    strat = AsyncPlaywrightCrawlerStrategy(browser_config=bc, browser_adapter=adapter)
    async with AsyncWebCrawler(crawler_strategy=strat, config=bc) as c:
        await c.crawler_strategy.browser_manager.start()
        return type(c.crawler_strategy.browser_manager.playwright).__module__

print("default :", asyncio.run(driver(PlaywrightAdapter())))   # playwright.async_api._generated
print("undetect:", asyncio.run(driver(UndetectedAdapter())))   # patchright.async_api._generated
```

The provider as shipped takes the first (plain-playwright) path.

### Expected vs actual

- **Expected:** with patchright installed, fetches run through the stealth driver.
- **Actual:** fetches run through plain Playwright; passive bot fingerprinting is more likely to serve interstitials/empty pages.

### Fix

Build the crawler with an explicit `AsyncPlaywrightCrawlerStrategy(browser_config=..., browser_adapter=UndetectedAdapter())` and pass it as `crawler_strategy=`. crawl4ai threads `use_undetected` through every `BrowserManager` launch branch (default, persistent-context `user_data_dir`, managed-browser/CDP), so it's compatible with the authenticated-profile path. `UndetectedAdapter`/`browser_adapter` was introduced in crawl4ai 0.7.3, so the dependency floor (currently `>=0.4`) should be bumped. PR incoming.

### Environment

- hyperresearch 0.8.6
- crawl4ai 0.9.0
- patchright (chromium installed)
- Python 3.11, macOS

seanyoungberg:
Fixes #35.

### Problem

`Crawl4AIProvider` built `AsyncWebCrawler(config=...)` with no `crawler_strategy`, so crawl4ai used the default `PlaywrightAdapter` (plain Playwright) and patchright's stealth never engaged — the "stealth" was only `simulate_user` + smart-wait.

### Change

- Add a small `_make_crawler()` helper that builds an `AsyncWebCrawler` wired with an explicit `AsyncPlaywrightCrawlerStrategy(browser_adapter=UndetectedAdapter())`.
- Use it at both browser call sites (`_fetch_async`, `_fetch_many_async`).
- Bump the `Crawl4AI` floor `>=0.4` → `>=0.7.3` (the release that introduced `crawl4ai.browser_adapter` / `UndetectedAdapter`; the new import would otherwise fail on the old floor).
- Add offline tests for the wiring.

### Why it's safe with profiles

crawl4ai threads `use_undetected=isinstance(adapter, UndetectedAdapter)` into `BrowserManager`, which selects patchright across all launch branches (default, persistent-context `user_data_dir`, and managed-browser/CDP). The authenticated-profile path keeps working — covered by `test_undetected_wiring_survives_profile_path`.

### Reproduction (before / after)

Runtime driver module backing the crawler:

```
before (default adapter) -> playwright.async_api._generated   (plain Playwright)
after  (UndetectedAdapter) -> patchright.async_api._generated (stealth)
```

End-to-end, the patched provider renders a JS-heavy passive bot detector with webdriver signals suppressed (`webDriver` / `webDriverAdvanced` report "OK"). Note: headless still exposes `HeadlessChrome` in the UA — patchright-headless reduces passive detection but does not defeat interactive Cloudflare Turnstile challenges; that's unchanged and out of scope here.

### Tests

`tests/test_web/test_crawl4ai_stealth.py` (offline, no browser launch):
- `test_make_crawler_uses_undetected_adapter`
- `test_browser_manager_marked_undetected`
- `test_undetected_wiring_survives_profile_path`

### Verification

- `ruff check src/` → clean
- `ruff format --check` on the touched files → clean (repo-wide format left untouched to keep the diff focused)
- `python -m pytest tests/` → all pass

The diff is intentionally minimal — only the two call sites, the helper, the dependency floor, and the new test.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     "pydantic>=2.0",
     "jinja2>=3.1",
     "platformdirs>=4.0",
-    "Crawl4AI>=0.4",
+    "Crawl4AI>=0.7.3",
     "pymupdf>=1.24",
     "httpx>=0.27",
 ]
--- a/src/hyperresearch/web/crawl4ai_provider.py
+++ b/src/hyperresearch/web/crawl4ai_provider.py
@@ -15,6 +15,8 @@ import sys
 from datetime import UTC, datetime
 
 from crawl4ai import AsyncWebCrawler, BrowserConfig, CrawlerRunConfig, DefaultMarkdownGenerator
+from crawl4ai.async_crawler_strategy import AsyncPlaywrightCrawlerStrategy
+from crawl4ai.browser_adapter import UndetectedAdapter
 from crawl4ai.content_filter_strategy import PruningContentFilter
 
 from hyperresearch.web.base import WebResult
@@ -195,6 +197,23 @@ class Crawl4AIProvider:
             markdown_generator=self._md_generator,
         )
 
+    def _make_crawler(self) -> AsyncWebCrawler:
+        """Build an AsyncWebCrawler wired with the stealth (patchright) adapter.
+
+        crawl4ai's AsyncWebCrawler defaults to PlaywrightAdapter (plain
+        playwright); patchright/stealth only engages when an UndetectedAdapter is
+        passed via an explicit AsyncPlaywrightCrawlerStrategy. Without this the
+        provider's anti-bot behavior is only simulate_user + smart-wait. The
+        adapter's use_undetected flag is threaded through every BrowserManager
+        launch branch (default, persistent-context, and managed-browser/CDP), so
+        this is compatible with the authenticated-profile (user_data_dir) path.
+        """
+        strategy = AsyncPlaywrightCrawlerStrategy(
+            browser_config=self._browser_config,
+            browser_adapter=UndetectedAdapter(),
+        )
+        return AsyncWebCrawler(crawler_strategy=strategy, config=self._browser_config)
+
     def fetch(self, url: str) -> WebResult:
         # PDF detection: fetch directly with httpx, extract text with pymupdf
         if _is_pdf_url(url):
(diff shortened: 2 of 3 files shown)


jordan-gibbs (approved):
The diagnosis checks out. I traced the default wiring and `AsyncWebCrawler` does fall back to plain `PlaywrightAdapter`, so stealth never engages. I also verified that crawl4ai 0.7.3+ declares patchright as a hard dependency, which means the floor bump covers the runtime import too.

The offline tests pin `browser_manager.use_undetected`, which is the flag that crawl4ai actually checks at launch. Tests pass locally too. This will need a small rebase once #41 and #42 land (all three touch `crawl4ai_provider.py`).

PR #40, opened 2026-07-18
fix: make CI install what the tests import, and lock packaging metadata

synqing:
## Why

`main` has failed CI since `c9abce7` (2026-05-14). The Exa provider tests import `exa_py`, but CI installs only `.[dev]`, which doesn't include it — so `tests/test_web/test_exa_provider.py` fails 10 of its 12 tests with `ModuleNotFoundError: No module named 'exa_py'` ([run 25879578650](https://github.com/jordan-gibbs/hyperresearch/actions/runs/25879578650)). The other 2 pass because they mock the `ImportError` path.

While fixing that, two smaller user-facing issues surfaced:

**The published wheel misreports its own version.** `pip install hyperresearch==0.8.6` gives you a CLI that prints `hyperresearch v0.8.5`, because `__init__.py` was left at `0.8.5` when `pyproject.toml` moved to `0.8.6`. This is live on PyPI today.

**`pip install hyperresearch[watch]` is a dead end.** The `watch` extra was never declared and `watchdog` is in no dependency list, so that command matches no extra and installs nothing — then `hpr watch` fails with an ImportError whose message tells you to run the command that just failed (`src/hyperresearch/cli/watch.py:21`). Same for `hyperresearch[crawl4ai]`, referenced in three places (`web/base.py:162`, `cli/setup.py:67`, `cli/install.py:172`), though that one is harmless since `Crawl4AI` is already a base dependency.

## What changed

- **`dev` extra now includes `exa-py`**, so `pip install -e ".[dev]"` — the exact command CI runs — installs what the suite imports. Also added `mcp`, `watchdog`, and `types-PyYAML`.
- **Declared the `crawl4ai` and `watch` extras** the CLI already directs users to.
- **Fixed the version skew**: both `pyproject.toml` and `__init__.py` now read `0.8.6`.
- **Added `tests/test_core/test_packaging.py`** so both failure modes stay fixed: runtime `__version__` must track `pyproject.toml`, and `dev` must cover the optional providers the suite imports.
- **`publish.yml` now runs ruff and pytest before building**, so a tagged release can't reach PyPI without passing the same checks as CI. `ci.yml` also proves the package still builds.

## On versioning

I deliberately **did not bump the version** — that's your call. The changelog entry sits under `## [Unreleased]` so you can rename it to whatever you cut next.

## Verification

Ran your CI workflow locally against the full declared support matrix (`>=3.11,<3.14`):

| Python | install | ruff | pytest | build |
|---|---|---|---|---|
| 3.11.13 | pass | pass | 193 passed | pass |
| 3.12.11 | pass | pass | 193 passed | pass |
| 3.13.11 | pass | pass | 193 passed | pass |

The 12 Exa tests are collected and executed, not skipped.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-      - run: pip install -e ".[dev]"
+      - run: pip install -e ".[dev]" build
       - run: ruff check src/ tests/
       - run: pytest tests/ -v
+      - run: python -m build --outdir /tmp/hyperresearch-dist
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,8 +21,14 @@ jobs:
         with:
           python-version: "3.13"
 
-      - name: Install build tools
-        run: pip install build
+      - name: Install test and build tools
+        run: pip install -e ".[dev]" build
+
+      - name: Lint
+        run: ruff check src/ tests/
+
+      - name: Test
+        run: pytest tests/ -v
 
       - name: Build package
         run: python -m build
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [Unreleased]
+
+### Release readiness and deployability
+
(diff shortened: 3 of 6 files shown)


jordan-gibbs (approved):
I ran everything locally on 3.11 and it's all green - ruff is clean and 193/193 pass. This one goes in first so the other PRs have a working CI to land against.

The version skew on the live 0.8.6 wheel is embarrassing (the wheel says 0.8.6, the CLI says 0.8.5). Thanks for catching it!

Issue #32, opened 2026-06-13
Empty-string query to 'hpr search' silently returns [] (FTS5 syntax error swallowed)


ankaggarwal94:
## Summary

An empty (or whitespace-/punctuation-only) query passed to `hpr search` silently returns `[]` instead of listing notes or raising an error. The empty query is turned into an empty FTS5 `MATCH` string, which SQLite rejects as a syntax error, and that error is swallowed by a bare `except`. This is a problem because **several of the shipped pipeline skill/agent prompts use `search "" --tag <tag>` as a "list every note with this tag" idiom**, so those vault surveys silently see zero notes.

## Environment

- hyperresearch **0.8.6** (installed via pipx); reproduced against `main` HEAD as well (`v0.8.6`)
- Python **3.13**
- macOS — but this is platform-independent (it's a SQLite FTS5 query-grammar issue, not OS-specific)

## Expected vs. actual

- **Expected:** `hpr search "" --tag mytag -j` either lists all notes carrying `mytag`, or fails loudly with a clear error.
- **Actual:** it returns an empty JSON array `[]` and exits 0 — indistinguishable from "no matching notes," even when many tagged notes exist.

## Steps to reproduce

The mechanism reproduces directly against the project's own `preprocess_query` plus an in-memory FTS5 table (no vault required):

```python
import sqlite3
from hyperresearch.search.fts import preprocess_query

print(repr(preprocess_query("")))      # -> ''   (also '   ' and '()' -> '')

con = sqlite3.connect(":memory:")
con.execute("CREATE VIRTUAL TABLE t USING fts5(body)")
con.execute("INSERT INTO t(body) VALUES ('hello world')")
con.execute("SELECT * FROM t WHERE t MATCH ?", ("",)).fetchall()
# raises: sqlite3.OperationalError: fts5: syntax error near ""
```

CLI manifestation:

```
hpr search "" --tag <some-tag> -j     # -> []  (silently, even when tagged notes exist)
hpr note list --tag <some-tag> --all -j   # correctly lists them
```

## Root cause

`src/hyperresearch/search/fts.py`:

1. `preprocess_query("")` returns `""` (an empty string `.split()`s to `[]`, and `" ".join([])` is `""`).
2. That empty string is bound to `WHERE notes_fts MATCH ?`, which raises `sqlite3.OperationalError: fts5: syntax error near ""`.
[...]

PR #41, opened 2026-07-18
fix: stop discarding non-English pages and swallowing search errors (#37, #32)

### LINKED ISSUE #37: `looks_like_junk()` rejects all CJK / non-ASCII pages as "binary/non-printable content"
Every successfully-fetched Japanese-language page and PDF is discarded with:

```
JUNK_CONTENT: High ratio of binary/non-printable content
```

The content is **not** binary — crawl4ai/pymupdf extract perfectly good UTF-8 text. The junk heuristic in `looks_like_junk()` counts any character with a code point `> 127` as "non-printable." Since **every** CJK character (and all accented/non-Latin text) has a code point above 127, any page that is more than 15% non-ASCII trips the threshold and is thrown away. For Japanese prose, that is ~100% of the body, so it always fails.

This makes the tool effectively unusable for non-English research: in a real session, *all* government sources (国税庁/NTA, 公正取引委員会/JFTC, METI, 文化庁, 個人情報保護委員会/PPC), every JP vendor page, and every JP-language PDF were dropped as "junk," even though they fetched and decoded fine.

Proposed fix:
Make `looks_like_junk()` use the same garbage definition already proven in `crawl4ai_provider._looks_like_binary()` — count only true control chars (`< 0x20` excluding whitespace), the Unicode replacement char `U+FFFD`, and C1 controls (`0x80–0x9F`):

synqing:
Fixes #37 and #32. Both are silent-failure bugs in the layers the vault's value rests on — neither surfaces an error, so you only notice by discovering content that should be there isn't.

## #37 — every non-English page is discarded as binary garbage

`WebResult.looks_like_junk()` (`web/base.py`) counted every `ord(c) > 127` character as non-printable and rejected content above a 15% ratio:

```python
non_printable = sum(1 for c in sample if ord(c) > 127 or c in '\x00\x01\x02\x03\x04\x05')
if non_printable > len(sample) * 0.15:
    return "High ratio of binary/non-printable content"
```

Chinese, Japanese, Korean, Arabic, Hebrew, Cyrillic, Greek, and Thai text is ~100% non-ASCII, so it always trips this. Those pages are fetched successfully and then thrown away.

**The correct check already existed in this repo** — `crawl4ai_provider._looks_like_binary()` has a comment spelling out exactly why `ord(c) > 127` is wrong. It just never got applied to the second gate, and the two copies drifted.

Fixed by extracting one shared `binary_garbage_ratio()` into `base.py` — counting only C0/C1 control characters and U+FFFD — and having both call sites use it, so they can't drift apart again.

Measured against the old predicate on real prose fixtures:

| fixture | old verdict | new garbage ratio |
|---|---|---|
| Chinese | REJECTED | 0.000 |
| Japanese | REJECTED | 0.000 |
| Arabic | REJECTED | 0.000 |
| Russian | REJECTED | 0.000 |

Genuinely binary content, mojibake (U+FFFD), and raw PDF internals are all still rejected — there are tests for each.

## #32 — a broken index is indistinguishable from an empty topic

`search_fts()` wrapped its query in a bare handler:

```python
except sqlite3.OperationalError:
    return []
```

That swallows an empty query, a malformed query, a missing `notes_fts` table, and a corrupt index alike — all reported as "no results". The empty-query case in #32 is the visible symptom; the dangerous one is a damaged vault silently reporting that you never wrote anything on the topic.

Now:
- degenerate queries raise `SearchQueryError` naming the reason
- genuine operational failures propagate instead of being masked
- valid queries matching nothing still return `[]`, unchanged

Callers in the CLI, MCP server, and web UI handle it rather than tracebacking. The CLI exits 2 and emits a `BAD_QUERY` code in `--json` mode. The web UI's existing empty-string guard didn't cover degenerate input like `***`, so that path is guarded too.

## Notes

- Ruff clean. Locally 210 pass / 0 fail.
- CI on this branch will still show the 10 pre-existing `exa_py` failures that have had `main` red since `c9abce7` — unrelated to this change, fixed separately in #40.
- No version bump; changelog left alone so you can group this into whatever you cut next.

### DIFF
--- a/src/hyperresearch/cli/search.py
+++ b/src/hyperresearch/cli/search.py
@@ -36,7 +36,7 @@ def search(
     from hyperresearch.core.vault import Vault, VaultError
     from hyperresearch.models.note import ContentType, Tier
     from hyperresearch.search.filters import SearchFilters
-    from hyperresearch.search.fts import search_fts
+    from hyperresearch.search.fts import SearchQueryError, search_fts
 
     # Validate enums up-front
     if tier is not None:
@@ -97,7 +97,16 @@ def search(
         "penalize_deprecated": vault.config.search_penalize_deprecated,
         "penalize_stale": vault.config.search_penalize_stale,
     }
-    results = search_fts(vault.db, query, filters=filters, limit=limit, offset=offset, ranking=ranking)
+    try:
+        results = search_fts(
+            vault.db, query, filters=filters, limit=limit, offset=offset, ranking=ranking
+        )
+    except SearchQueryError as e:
+        if json_output:
+            output(error(str(e), "BAD_QUERY"), json_mode=True)
+        else:
+            console.print(f"[red]Error:[/] {e}")
+        raise typer.Exit(2) from e
 
     # Auto-include body in JSON mode (agents almost always need it)
     want_body = include_body or (json_output and not no_body)
--- a/src/hyperresearch/mcp/server.py
+++ b/src/hyperresearch/mcp/server.py
@@ -43,7 +43,7 @@ def search_notes(query: str, tag: str = "", status: str = "", parent: str = "", 
     vault = _get_vault()
     vault.auto_sync()
     from hyperresearch.search.filters import SearchFilters
-    from hyperresearch.search.fts import search_fts
+    from hyperresearch.search.fts import SearchQueryError, search_fts
     tags = [t.strip() for t in tag.split(",") if t.strip()] or None
     filters = SearchFilters(tags=tags, status=status or None, parent=parent or None)
     ranking = {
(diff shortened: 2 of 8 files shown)


jordan-gibbs (approved):
I checked the claim against the source before I read your fix, and it holds up. `ord(c) > 127` was throwing away pretty much every non-English page. Moving the check into one shared `binary_garbage_ratio()` is the right fix, since the two gates can't drift apart again. The fixtures use actual prose in each language instead of synthetic strings.

I also confirmed that all three `search_fts` call sites handle `SearchQueryError`, and the full suite is green locally.

One heads-up is that the shipped skill prompts used `search "" --tag` as a list-all idiom, which this PR turns into a loud error. I'm landing a follow-up that rewrites those callers to `note list --tag --all`, so it's not a blocker for this.

Issue #39, opened 2026-07-11
`fetch` returns JUNK_CONTENT for direct PDF URLs across domains, even when the PDF has extractable text


mcowan38:
**Version:** hyperresearch 0.8.6 (pip). Note: the 0.8.6 wheel ships with `__version__` still at "0.8.5", so `--version` under-reports; `pip3 show hyperresearch` confirms 0.8.6.

**Environment:** Debian bookworm container (arm64), Python 3.11, Crawl4AI browser lane on Playwright-installed Chromium.

## What happens

During a full research run, every fetch of a **direct PDF URL** came back classified `JUNK_CONTENT`, across unrelated domains — observed against NBER, Harvard, and Cambridge (cambridge.org) PDF links. The PDFs themselves are fine: they download normally and have an extractable text layer (verified manually with the same URLs).

The pipeline routes around it — fetcher agents fall back to HTML abstract/landing pages, so topical coverage holds — but `research/raw/` ends up with **zero archived PDFs**, and every deposited reference then fails any downstream full-text/PDF-presence check. In our run, 16 of 16 references flagged missing-PDF for this reason alone.

## Expected

Direct PDF URLs with extractable text should be fetched and stored (or at minimum classified as fetchable binary content), not discarded as junk.

## Suspicion

The junk-content heuristic looks tuned for HTML and appears to run on the raw PDF bytes (or a failed text extraction of them), so any `application/pdf` response scores as garbage. A content-type gate before the junk classifier (route `application/pdf` to PDF extraction instead) would likely fix it.

## Repro

1. `hyperresearch fetch <any direct NBER/Cambridge PDF URL>` (e.g. an `nber.org/system/files/working_papers/...pdf` link)
2. Observe `JUNK_CONTENT` result; nothing lands in `research/raw/`.
3. `curl -o x.pdf <same URL> && pdftotext x.pdf -` extracts text fine.

## Impact

Silent degradation: reports cite abstract-page content while the primary PDFs are never archived, which breaks downstream verification tooling that checks for full-text presence.

Possibly related: #37 (same junk classifier rejecting non-HTML content, there triggered by CJK pages).

jordan-gibbs:
[...]

PR #42, opened 2026-07-18
fix: make PDF fetch failures diagnosable instead of silent (#39)

synqing:
Addresses #39.

## What I found

I could not reproduce #39 against the reported domains. Fetching the NBER and Cambridge URLs directly returns HTTP 200 with `content-type: application/pdf` and valid `%PDF-` magic bytes, and `_fetch_pdf()` extracts them fine — 67 and 7 pages respectively, both passing `looks_like_junk()`.

I also checked the reporter's guess that this shares a root cause with #37 (the `ord(c) > 127` junk check). It doesn't: English academic PDFs are only 0.1–0.4% non-ASCII, well under that 15% threshold, so they were never rejected by it.

What does match the reported shape — *every* direct PDF URL failing across *unrelated* domains, on an arm64 container — is the dependency gate at the top of `_fetch_pdf`:

```python
try:
    import pymupdf
except ImportError:
    return None
```

If pymupdf has no working wheel for the platform, PDF extraction is disabled entirely and silently. Every PDF falls through to the browser lane, arrives as binary, and is discarded as junk — which presents as a fetching bug rather than a missing dependency. Nothing in the logs distinguishes the two.

I can't confirm that is what happened on the reporter's box. What I can do is make the failure impossible to misread next time, and fix the real gaps I did find.

## Changes

- **Missing pymupdf logs an ERROR naming the consequence** ("PDF text extraction is disabled, so every PDF will be discarded as junk content") plus the fix. Once per process, not once per URL.
- **Every other failure path logs its specific reason** — non-200 status, non-PDF body (with content-type and first bytes), undersized response, and no-text-layer. A scanned PDF now says it needs OCR instead of vanishing.
- **PDF identity now comes from `%PDF-` magic bytes**, not the content-type header or a `.pdf` URL suffix. Servers mislabel PDFs as `application/octet-stream`, and plenty of PDF URLs carry no suffix — the Cambridge URL in #39 is one, and `_is_pdf_url()` returns False for it, so it only reaches PDF extraction via the post-fetch binary retry.
- **Completed the User-Agent.** The old string was truncated mid-token (`...AppleWebKit/537.36`, no Chrome/Safari suffix), which is a known bot signature. To be clear: measured against live NBER and Cambridge URLs this was **not** the cause here — both return 200 with either UA — but a malformed UA is worth fixing on its own.

## Tests

`tests/test_web/test_pdf_diagnostics.py` — offline, httpx stubbed, no network. Covers each diagnostic path, plus a genuine PDF mislabelled as octet-stream (must be accepted) and a scanned PDF with no text layer (must explain itself). 209 pass, ruff clean.

If the reporter can share whether `python -c "import pymupdf"` succeeds in that container, that would confirm or eliminate the hypothesis directly.

### DIFF
--- a/src/hyperresearch/web/crawl4ai_provider.py
+++ b/src/hyperresearch/web/crawl4ai_provider.py
@@ -9,6 +9,7 @@ Supports authenticated crawling via crawl4ai browser profiles:
 from __future__ import annotations
 
 import asyncio
+import logging
 import os
 import pathlib
 import sys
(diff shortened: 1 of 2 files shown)


jordan-gibbs (approved):
Since nobody could reproduce the original report, logging why each fetch failed makes more sense than guessing at a fix. The pymupdf import gate also matches the symptom in the report, where every PDF on every domain fails at once, better than the junk-classifier theory did.

The magic-bytes check is a fix in its own right, since mislabelled octet-stream PDFs were getting dropped and HTML served with a PDF content-type now gets a clear log line. The full suite passes locally. I'll land this after #41 and #36.

PR #52, opened 2026-07-20
fix: close four silent-failure leaks (tags, FTS syntax, batch PDFs, cache-busting date)

fduple:
Four small, independent fixes for failures that produced wrong-but-
plausible results instead of errors:

- core/sync.py: lowercase frontmatter tags before the alias lookup and
  before INSERT (alias keys are stored lowercase, and SearchFilters
  lowercases the query side). Previously an uppercase tag missed its
  alias mapping and tag-filtered searches silently returned nothing.
  Regression tests in tests/test_core/test_sync.py, including the
  aliased-uppercase case.

- search/fts.py: strip stray unbalanced `"` from query terms during
  preprocessing — it previously raised an FTS5 syntax error that the
  caller saw as zero results. Hyphenated words are left intact: bare
  tokens are emitted inside double quotes, where `-` is a plain
  character, so `foo-bar` stays a phrase query. Regression tests in
  tests/test_search/test_fts.py.

- web/crawl4ai_provider.py: in _fetch_many_async, PDFs that fail direct
  fetch fall back to the browser lane (mirrors single-fetch behaviour).
  Batch callers no longer silently lose academic PDFs behind
  JS-rendered landing pages.

- cli/fetch_batch.py: when the batch call fails wholesale, fall back to
  a per-URL loop, and surface `failed_urls` in the JSON envelope so the
  caller can see exactly what was lost instead of inferring it from
  count mismatches.

- core/agent_docs.py: drop the `Today is {date}` line from the injected
  CLAUDE.md blurb — a date in the cached prefix busts Claude Code's
  prompt cache once per day for every project the blurb is installed in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Four silent-failure leaks closed (tags, FTS syntax, batch PDFs, cache-busting date)
+
+- **Tag filtering:** frontmatter tags are now lowercased before the alias lookup and before storage, matching the lowercase alias keys and the lowercasing `SearchFilters` does on the query side — `--tag llm` now finds notes tagged `LLM`, including through aliases.
+- **FTS queries:** stray unbalanced double-quotes in bare words are stripped instead of raising an FTS5 syntax error that the caller saw as zero results. Hyphenated words are left intact — they were already valid inside the emitted quoted phrase tokens.
+- **Batch PDF fetches:** in `fetch_many`, a PDF whose direct download fails (parser failure, non-PDF body at a `.pdf` URL) now falls back to the browser path, mirroring single-fetch behaviour instead of being silently dropped. `fetch-batch` also falls back to per-URL fetches when a whole batch fails, and reports `failed_urls` in its output instead of losing them silently.
+- **Prompt-cache-busting date:** the vault CLAUDE.md blurb no longer interpolates `Today is YYYY-MM-DD`, which busted Claude Code's prompt cache once per day.
+
 ## [0.9.0] - 2026-07-23
 
 ### Coverage before elegance: the synthesizer stops trading substance for prose
--- a/src/hyperresearch/cli/fetch_batch.py
+++ b/src/hyperresearch/cli/fetch_batch.py
@@ -94,6 +94,15 @@ def fetch_batch(
     visible_urls = [u for u in new_urls if _needs_visible(u)]
 
     results = []
+    failed_urls: list[dict] = []  # surfaced in JSON output so callers see what was lost
+
+    def _fetch_one(provider, url: str, kind: str) -> None:
+        try:
+            results.append(provider.fetch(url))
+        except Exception as exc:
+            failed_urls.append({"url": url, "error": str(exc), "phase": kind})
+            if not json_output:
+                console.print(f"  [red]Failed ({kind}):[/] {url} — {exc}")
 
     # Batch fetch normal URLs
     if normal_urls:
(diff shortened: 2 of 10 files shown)


jordan-gibbs:
Thanks for these! The batch PDF catch is one I'm glad to have, I hadn't noticed we were dropping those.

A couple things I'd ask before I merge:

1. The tag fix lowercases after the alias lookup, but alias keys are stored lowercase (cli/tag.py:56), so an uppercase frontmatter tag still misses its mapping. `alias_map.get(tag.lower(), tag.lower())` gets both.
2. Can you drop the hyphen replacement in fts.py? Bare tokens already get quoted (fts.py:58), so foo-bar was a valid phrase query all along, and the replacement loosens phrase matching into AND'd tokens. The stray-quote strip is the part I want.

Small stuff: a CHANGELOG entry under Unreleased (house habit), and I think the "Review guide" paragraph in your description was meant for #51.

fduple:
Thanks for the sharp catches. Both were exactly right.

- Tag fix: now `alias_map.get(tag.lower(), tag.lower())`, with a new test
  covering the aliased-uppercase case (`ML` resolving through an `ml ->
  machine-learning` mapping).
- Hyphen replacement dropped. The emitted tokens are already quoted, so
  `foo-bar` was a valid phrase query all along, and the replacement was
  loosening phrase matching for nothing. The stray-quote strip stays, and
  the tests now pin the phrase behaviour instead.
- CHANGELOG entry added under Unreleased.
- Good eye on the description. That "Review guide" paragraph was pasted
  from #51 by mistake; I've removed it.

Two follow-up commits since your review: the fixes above, then a separate
`tests:` commit adding offline coverage for the two web-layer changes this
PR introduced (the batch-PDF -> browser fallback and `fetch-batch`'s
`failed_urls`), which had no tests before.

fduple:
Rebased onto current main (v0.9.0) so it merges cleanly. No code changes since your review; I just moved my CHANGELOG entry under the now-empty Unreleased section after the release cut.

jordan-gibbs:
CI's red on the two new tests in test_fetch_many_fallback.py: `asyncio.run()` throws "cannot be called from a running event loop" in the full suite (passes in isolation, so it only shows up on CI). The fallback fix itself is fine, it's just the test driving the coroutine.

Quickest fix is to not assume there's no loop running:

    from concurrent.futures import ThreadPoolExecutor
    def _run(coro):
        with ThreadPoolExecutor(1) as ex:
            return ex.submit(lambda: asyncio.run(coro)).result()

then `_run(inst._fetch_many_async([...]))` instead of asyncio.run directly. Same for the mirror test.

Do that plus move the CHANGELOG entry up to [Unreleased] and I'll merge.

fduple:
Heads-up on the earlier 3.11 CI failure: it was a flaky test on my side, not the change. test_fetch_many_fallback called asyncio.run(), which errors when a prior test in the run leaves crawl4ai's event loop in a running state (intermittent, and only visible on 3.11; my setup with 3.12 tolerates it). I moved that coroutine onto an isolated thread so it's immune to the leaked loop. No  production code changed, and it's green on 3.11 and 3.12 now.

Tthe CHANGELOG entry looks good to me though, between [Unreleased] (line 3) and  [0.9.0] (line 12)

jordan-gibbs:
Thanks for turning these around. The CHANGELOG ask was me reading a stale diff, your entry is already under `[Unreleased]` where it belongs, so sorry for the runaround.

Fork PRs don't get CI here, so I ran it locally against current main and it's ruff clean with 562 passing, including the thread-isolated `_run` helper on the fetch_many tests. Squash-merging now.

The batch PDF catch is the one I'm most glad to have, that was quietly costing me sources on every wave.

PR #48, opened 2026-07-19
feat: 2.0 groundwork (profiles, source ranking, runs, browser lane, verification)

jordan-gibbs:
## What this is

The five-phase 2.0 roadmap (`docs/roadmap-2.0/`) implemented end to end, plus the scale-profile system and per-agent model flexibility built on top of it. The CHANGELOG `[Unreleased]` sections are the detailed narrative; the short version:

### Phase 1: config + profiles
Every behavioral constant moved into `.hyperresearch/config.toml`. Pipeline profiles (`[profile.*]`, `extends`) with Jinja prompt templating (`<< p.* >>`), strict rendering, and golden-fixture drift protection. A profile change that alters a prompt is a reviewed diff, never silent.

### Phase 2: source ranking
Source quality is persistent, not vibes: composite `quality_score` from tier, fetch-time utility, citation authority (OpenAlex / Semantic Scholar, with retraction flags), and vault PageRank. Claims table plus optional embeddings/semantic search (zero API keys by default).

### Phase 3: runs at scale
Per-run workspaces (`research/runs/<tag>/`), run manifest with `run init/status/resume/report/verify`, budget governor, and the chaptered `dissertation` profile (step 1.5, 300-450 sources).

### Phase 4: browser lane
Blocked fetches queue as escalations instead of dying; a browser-fetcher agent drains them via Claude-in-Chrome. Hard boundary: CAPTCHAs, 2FA, and logins always hand off to the human.

### Phase 5: verification
Cite-check step 14.5 (per-citation binding audit), `quote-integrity` / `numeric-consistency` / `retracted-citations` lints, independence audit (syndicated copies count as one voice), per-step run telemetry.

### On top

- **Scale profiles**: new `premier` profile (100-130 sweep targets; 250+ sources landed in practice); `hyperresearch profile use <name>` re-renders installed prompts and persists the choice.
- **Per-agent models are real config**: agent `model:` frontmatter renders from the profile. `models = { fetcher = "haiku" }` works now. Model names removed from all prose so overrides can never be contradicted by stale text.
- **No dollar-cost claims in the product**: estimates removed; spend telemetry relabeled API-equivalent. Enforced by tests.

## Numbers

18 step skills, 17 agents, **506 tests passing**, ruff clean. 102 files, +12,889/-629.

## Validation

Independently test-verified; first end-to-end premier run (bench query #62) is in flight. Release/version bump intentionally deferred until its report and RACE score land.

PR #53, opened 2026-07-20
security: SSRF gate + size caps for fetches (builtin, crawl4ai entry points, PDFs, images)

fduple:
Problem: every outbound fetch path (builtin._download, the image downloader, crawl4ai _fetch_pdf, Crawl4AIProvider.fetch) would happily request any URL a fetched page or search result handed it. An attacker page could point a fetcher at 169.254.169.254 (cloud metadata), 127.0.0.1, or RFC1918 space — or stream an unbounded body and OOM the worker.

Fix: new web/safe_http.py provides safe_get():
- http/https schemes only; hostname resolved and refused if private, loopback, link-local, multicast, reserved, or unspecified;
- redirects followed MANUALLY, each hop re-validated against the gate;
- response streamed into a fixed cap (MAX_BYTES_HTML/PDF/IMAGE) so a hostile server cannot exhaust memory;
- returns a small SafeResponse dataclass, no httpx types leaked.

Wired into builtin._download, the image downloader (honoring the configured asset timeout), _fetch_pdf (honoring pdf_timeout_s), and as a check_url() entry gate in Crawl4AIProvider.fetch (Playwright drives its own navigation, so the entry check is the browser path's guard).

TLS policy, composing with the pdf_verify_tls setting introduced in 2.0 groundwork (#48): verified-first with one unverified retry. With pdf_verify_tls = true (the default), certificate verification is on; if and only if a fetch fails with a certificate error *(common on academic PDF hosts with expired/misconfigured certs),* one retry runs with verification off, logged loudly ("tls-unverified fetch") for provenance — strictly more reach than fail-closed, strictly more integrity than verify-off. An explicit pdf_verify_tls = false keeps its documented meaning: unverified from the first attempt.

Tests: tests/test_web/test_safe_http.py (scheme/IP refusal, redirect re-validation, size caps, header handling) plus three regression tests pinning retry-once, no-retry-on-other-errors, and the config opt-out. The test_pdf_diagnostics stub now fakes the gated download layer instead of httpx.get, so the diagnostics tests exercise the real code path.


jordan-gibbs:
Thanks for taking this on, SSRF was on my list and I hadn't gotten to it.

There's a blocker, and a few things I'd ask of you.

The blocker: the unverified TLS retry needs to go, or become a new opt-in setting that's off by default. Any MITM can hand back a bad cert to force the retry, which flips my verify-by-default decision (config.py:16) into something an attacker controls, and those unverified bytes go straight into pymupdf. pdf_verify_tls = false already covers known-broken mirrors.

The asks:

1. The gate resolves the hostname but httpx re-resolves on connect, so low-TTL rebinding slips through. Pin the validated IP, or soften the docstring to best-effort.
2. fetch_many isn't gated, and that's the main fetch path, so "all outbound fetches" isn't true yet.
3. Redirect revalidation and the size caps have no tests.
4. Make the caps FetchSettings fields, and _download_image swallows the new refusals silently.
5. CHANGELOG entry.

Once the TLS retry is sorted I'm happy to merge and let the rest come as follow-ups if you'd rather split it.

fduple:
Agreed on the blocker. The retry was exactly the kind of "attacker
controls the downgrade" hole you described, and `pdf_verify_tls = false`
already covers the legitimate case. It's gone: a certificate failure now
refuses with a message that names the config opt-out, and the old retry
tests were rewritten to pin the no-retry behaviour.

The asks, all addressed on top:

1. Rebinding: I went with softening rather than pinning. httpx re-resolves
   at connect time, and pinning the validated IP would need a custom
   transport. The module docstring now states the check is best-effort and
   documents the residual window explicitly.
2. `fetch_many` is gated with the same `check_url` as `fetch()`; refused
   URLs in a batch are logged and skipped rather than killing the wave.
   There's a test for it.
3. Tests added for redirect revalidation (public→private bounce, relative
   redirects, redirect loops) and both size-cap paths (Content-Length
   refusal, and a mid-stream cut for chunked bodies with no length header).
   They run offline via an httpx.MockTransport seam on safe_get.
4. Caps are now `[fetch]` settings (`max_html_bytes` / `max_pdf_bytes` /
   `max_image_bytes`), threaded through the builtin provider, PDF and image
   paths; `_download_image` prints gate refusals instead of swallowing them
   with routine 404s/timeouts.
5. CHANGELOG entry added.

The encoded-loopback bypass and byte-cap threading tests are in a separate
`tests:` commit on top of the review.

fduple:
Follow-up force-push, test-only: hardened test_fetch_many_gate against a 3.11 flake. It used asyncio.run(), which errors when an earlier test in the run leaves crawl4ai's event loop running (intermittent, and only surfaces on 3.11; my local 3.12 tolerates it). It now runs the coroutine on an isolated thread, immune to the leaked loop. No production code changed; green on 3.11 and 3.12.

jordan-gibbs:
The TLS retry removal is what I wanted, and having the refusal name `pdf_verify_tls = false` is better than what I asked for.

I nearly signed this off thinking the browser path was an edge case. The `FetchSettings` default is `web_provider = "builtin"`, so I read the entry-point-only comment in `fetch()` as covering a minority of runs, and then I remembered that I wrote `cli/setup.py:63`, which asks "Use crawl4ai as the default web provider?" with `default=True`. My own vault config says crawl4ai. That dataclass default only applies when there's no config file at all, so the least-gated path in this PR is the one that pretty much everyone is running. That's on me rather than you, but it's the thing I want closed before this lands.

The cheap version is a recheck of the final URL after the fact. The browser follows redirects on its own, so the URL you validated at the entry point isn't necessarily where you ended up, and each path already has the final URL sitting there:

- `_fetch_async` has `result.url` after `arun`
- `_fetch_many_async` has `cr.url` in the zip loop, which you're already using on the `WebResult(url=cr.url or url)` line
- `_fetch_visible` has `final_url`, computed from `page.url`

Running `check_url` on those and refusing the result won't stop the request firing, so blind SSRF survives it, but it keeps private-network content out of the vault. If crawl4ai has a per-navigation hook that does this properly I'd rather have that, though I don't want to send you hunting for one.

The other thing I'd ask for is an escape hatch. `_is_blocked_ip` refuses every private and loopback address with no way out, so somebody running a self-hosted paper mirror or an intranet wiki on 10.x can't fetch it at all. Something like `[fetch] allow_private_hosts` that takes hostnames or CIDRs and gets consulted inside `check_url`, empty by default. That keeps the secure behaviour on a fresh install, and adding an entry stays an explicit act by someone who controls the address.

On rebinding, take the softening. An attacker has to already control DNS for a host I'm fetching, and I'd rather have the docstring being honest about the residual window than a custom transport that you'd have to hand-roll. Leave it as it is.

I'd change the wording too, since "all outbound fetches" isn't true yet. `core/embed.py:66,79` and `core/scholar.py:96` still go out on bare httpx, and I read through both, neither one is an SSRF vector since the hosts are fixed and scholar quotes the DOI with `quote(doi, safe='')`. So there's nothing to fix in the code, but the title and the CHANGELOG entry should say what's actually covered, which is builtin, the crawl4ai entry points, PDF downloads, and image downloads.

Small stuff: I checked on 3.11.9 and `100.64.0.0/10` (CGNAT) passes every `is_*` property, so it isn't blocked. One line in `_is_blocked_ip` if you want it. I'm not asking you to fix either of these here, but the cap bounds bytes and not time, so a server trickling a byte every 29 seconds stays under both the read timeout and the cap while it holds a fetcher open. And `check_url` in `_fetch_many_async` does a blocking `getaddrinfo` per URL on the event loop before any await, which I measured at about 2.5s on an 80-URL wave. Both can be follow-ups.

I ran it locally against current main and it's ruff clean with 602 passing. Once the final-URL recheck and the allowlist are in, I'll merge.

fduple:
Both asks are in, plus one thing your #52 merge surfaced when I rebased.

1. Final-URL recheck, your cheap version. All three landing points now
   re-run `check_url` on the URL the browser actually ended up on:
   `result.url` in `_fetch_async`, `cr.url` in the `_fetch_many_async`
   zip loop, and `final_url` in `_fetch_visible`. A single fetch that
   lands somewhere refused raises; a batch result is dropped with a
   warning while the rest of the wave survives. When the final host
   equals the entry host the recheck is skipped, so the common
   no-redirect case pays no second `getaddrinfo` (which also keeps this
   from making the event-loop blocking you measured any worse). I did
   not go hunting for a per-navigation hook, as you suggested; if
   crawl4ai grows one, the recheck can move into it later. The code
   comment and the CHANGELOG both say plainly that a recheck cannot
   stop the request that already fired, so blind SSRF survives it; what
   it buys is that private-network content stays out of the vault.

2. `[fetch] allow_private_hosts`, empty by default. Entries are either
   hostnames (exact match, case-insensitive) or CIDRs (a bare IP counts
   as its own /32), consulted inside `check_url`, so the allowlist works
   everywhere the gate does: entry checks, every `safe_get` redirect
   hop, and the new final-URL recheck. A malformed CIDR errors loudly
   naming the entry; silently ignoring it would present as an SSRF
   refusal of the exact mirror the user just tried to admit.

3. CGNAT one-liner taken: `100.64.0.0/10` is now refused, with a test
   pinning both /10 boundaries and their public neighbors.

4. Wording fixed. The CHANGELOG entry now says what is actually covered
   (builtin provider, crawl4ai entry points, PDF and image downloads)
   and names `core/embed.py` and `core/scholar.py` as deliberately out
   of scope since their hosts are fixed. I have retitled the PR to
   match.

5. Found during the rebase: your #52 merge sends a failed PDF to the
   browser lane, and on this branch a certificate refusal was one of
   the failures that returned None. Composed, that meant a cert-refused
   PDF got retried through a lane that runs with TLS errors ignored,
   which is the automatic unverified retry we just removed, wearing a
   different hat. Cert refusals are now their own exception type
   (`CertVerificationError`), propagate on the single-fetch path, and
   are a loud skip in a batch. They never reach the browser fallback.
   Tests pin both paths.

6. One correction to the plan we both signed off on, caught by driving
   a real chromium through a local 302 after the recheck was in:
   `CrawlResult.url` is the REQUESTED url even after the browser
   followed redirects, so rechecking `result.url` and `cr.url` was
   re-validating the entry URL — a no-op for exactly the redirect case
   it exists for. The landing url is `CrawlResult.redirected_url`; the
   recheck now consults that (getattr-guarded for older crawl4ai) on
   both headless lanes. `page.url` on the visible lane was already the
   real location. Stored `WebResult.url` behavior is unchanged. It's a
   separate commit on top so you can see exactly what moved after your
   review.

On your two follow-ups (a time bound to go with the byte cap, and moving
the batch `getaddrinfo` calls off the event loop): agreed on both, left
out of this round as you suggested.

Rebased onto current main (post-v0.9.1). Full suite and ruff green on
3.11 and 3.12.

jordan-gibbs:
Sorry this sat — it shouldn't have, given what it is.

I traced the exposure rather than taking the writeup on faith, and it's reachable three ways, one of which needs no model cooperation at all:

1. The fetcher agent's Phase 2 is mandatory, so it extracts URLs out of fetched page bodies and fetches them — and page bodies are already treated as hostile elsewhere in the tree.
2. `mcp/server.py:283` exposes `fetch_url` to any MCP client with no human in the loop.
3. `cli/fetch.py:756` calls `urllib.request.urlopen` on `` values harvested from the page, filtered only by `startswith("http")`, with an unbounded `read()`. An `` in the top-5 scored images fires on every `--save-assets` fetch, and a body over `min_image_bytes` gets written to `research/assets/` with its path returned in the output.

Response bodies become notes the agent reads, so this is read-SSRF, not blind.

I also checked the gate against the usual ways these go wrong: redirects are revalidated per hop (not just the first URL), IPv4-mapped and link-local forms are blocked, and decimal/octal encodings fail closed because you resolve through `getaddrinfo` rather than parsing. DNS rebinding stays open, and I'm happy with it being documented honestly rather than half-fixed.

CI is green on it. Three things before I merge:

- **Rebase.** The only conflict is the `CHANGELOG.md` stanza — your `[Unreleased]` block anchors on `## [0.9.1] - 2026-07-25` and main is past that now. Every source file still auto-merges; I checked each hunk's context. Head of main is `v0.10.1`.
- **`core/fetcher.py:176` passes `max_image_bytes` but not `allow_private_hosts`,** so the MCP and core paths ignore a user's allowlist. Fails closed, so it's not a hole, but it'll generate a confusing bug report.
- **6to4 (`2002::/16`) isn't blocked.** Exotic and deprecated, but it's a one-liner next to the CGNAT range you already handle.

One follow-up I'll take myself rather than ask of you: `core/oa.py:116` carries a narrow duplicate of this check with a comment saying to collapse it into `check_url` once this lands, and `oa._http_get_text` (added after you opened this) has a 60s timeout and no size cap. Both are mine to clean up. Worth keeping your credentials-in-URL rejection when they merge — `check_url` doesn't have it.

jordan-gibbs:
Merged. I applied the three follow-ups myself rather than round-trip them: rebased the CHANGELOG stanza, threaded `allow_private_hosts` through `core/fetcher.py`, and added 6to4 decoding (`2002::/16` embeds an IPv4 address that `ipaddress` doesn't flag — `2002:a9fe:a9fe::` was cloud metadata in an IPv6 coat). Tests for both. Thanks for the care on this one; the redirect revalidation and fail-closed resolution are exactly right.

PR #57, opened 2026-07-23
feat: add Parallel Search MCP provider

georgeatparallel:
## Summary

Add an opt-in `parallel` web provider backed by the anonymous Parallel Search MCP endpoint. It maps Hyperresearch's existing search and fetch interface to `web_search` and `web_fetch`, including response validation and endpoint input bounds.

The default provider and the existing crawl4ai, Exa, and Tavily paths are unchanged.

The provider now has its own optional dependency contract:

- `hyperresearch[mcp]` remains `mcp>=1.6,<2` for Hyperresearch's FastMCP server.
- `hyperresearch[parallel]` requires `mcp>=1.8,<2`, the first release with the Streamable HTTP client used by this provider.

The provider uses the backward-compatible `streamablehttp_client` spelling, accepts both older text-only and newer structured MCP results, applies `FetchSettings.page_timeout_ms` to transport, session, and tool-call reads, and sends one random process-level session ID for free-tier logging and rate-limit correlation.

## Install

```bash
pip install "hyperresearch[parallel]"
```

```toml
[web]
provider = "parallel"
```

## Testing

- Full test suite with the development environment
- Full test suite from the built wheel with MCP confirmed absent
- Focused provider and packaging tests
- Ruff lint across `src` and `tests`
- Ruff formatting for the changed provider and test files
- Strict mypy for the provider
- Wheel and sdist build, including optional-dependency metadata inspection
- Live anonymous search and fetch smoke tests with MCP 1.8.0 and MCP 1.29.0
- `git diff --check`

Disclosure: I work at Parallel.ai, which operates this MCP endpoint.


jordan-gibbs:
Thanks for the disclosure up front, and for making it opt-in with the default left alone.

The test module imports `mcp.types` at module scope, so `pytest tests/` dies at collection for anybody who doesn't have the extra installed. I hit it on a working checkout. The pattern I've been using is in `tests/test_web/test_tavily_provider.py`, which injects a fake module into `sys.modules` so the suite passes without tavily-python present, and its docstring explains why. Could you follow that one and drop the hard import? CI wouldn't have caught this, since the dev extra pulls mcp in.

On `mcp>=1.24,<2`, I went and looked at the wheels. The floor is genuinely forced by `streamable_http_client`, because the underscored spelling doesn't exist before 1.24, though `streamablehttp_client` is still exported in 1.28.1 and goes back past 1.12. What makes me hesitate is that `hyperresearch[mcp]` also installs the `hpr mcp` FastMCP server, so this floor forces an upgrade on people who took the extra for something that has nothing to do with your provider. If the older alias works for you I'd rather hold the floor at `>=1.6` and revisit it when I move the server. If you want 1.24 for a reason beyond the spelling, tell me and I'll take it as it is.

Small stuff: there's no timeout on the MCP session, so a hung endpoint stalls a fetch indefinitely, and the other providers thread `FetchSettings` timeouts through. There's also no `fetch_many`, so bulk waves fall back to the per-URL loop at `cli/fetch_batch.py:109` and open a fresh MCP connection for every URL. A run here routinely pulls 55-130 sources, so I think that one is worth a follow-up even if it isn't this PR.

Last thing, `session_id` is a stable per-process UUID that goes to your endpoint for free-tier correlation. That seems reasonable to me, but it should be visible to whoever's running it, and a line in the provider docstring saying requests within a process are correlated would cover it.

georgeatparallel:
Thanks for the detailed pass — I addressed each point in the rebased branch:

- The provider tests now inject fake `mcp`, `mcp.client`, and transport modules before importing the provider, following the Tavily pattern. The full suite also passes from the built wheel in an environment where MCP is confirmed absent.
- I kept the server-oriented `hyperresearch[mcp]` extra at `mcp>=1.6,<2` and added a separate `hyperresearch[parallel]` extra at `mcp>=1.8,<2`. I verified 1.8 is the first release with `streamablehttp_client`, and exercised it against the live endpoint. The provider uses that older alias and falls back cleanly when `structuredContent` is unavailable on older SDK results.
- `FetchSettings.page_timeout_ms` now applies to the MCP transport, SSE read, client session, and tool call.
- The session ID is now genuinely process-level, and both the provider docstring and README disclose that Parallel uses it for free-tier log correlation and rate limiting.
- The branch is rebased on current `main`; GitHub now reports it mergeable.

I left `fetch_many` for a focused follow-up, as you suggested, rather than expanding this PR's connection lifecycle and partial-failure semantics.

Validation included the full suite with and without MCP installed, Ruff, strict mypy, wheel/sdist builds, and live `web_search`/`web_fetch` smoke tests on MCP 1.8.0 and 1.29.0.

jordan-gibbs:
Thanks for disclosing the affiliation up front without being asked, and for actually addressing both of my earlier notes — the module-scope `mcp.types` import and the `mcp>=1.24` floor that would have dragged an unrelated upgrade onto everyone using `hyperresearch[mcp]`.

The code is well-behaved: its own optional extra, a single `elif` in `web/base.py:226` shaped like the existing exa and tavily branches, default provider untouched, and tests that stub the SDK so the suite still collects without the extra installed. And it fills a real gap — `builtin` raises `NotImplementedError` for search, so today the only working options both require a paid key.

Two changes and I'll merge:

**Move the README section out of the quickstart.** Right now it sits between the install command and the pipeline section. Neither exa nor tavily is mentioned anywhere in that file, so as it stands a vendor-authored provider gets placement no incumbent has. A providers section further down, alongside the other two, is the right home.

**Drop "free".** The README calls it "Parallel's free Search MCP endpoint" and references the free tier. I keep third-party pricing claims out of the repo on principle — I can't hold them stable, and when they change it's my README that's wrong. "Keyless" or "no account required" says the useful part without making a claim about your pricing.

Two things I'm accepting but want stated plainly in the thread rather than only in the code, so nobody's surprised later:

- Every request carries a process-level random `session_id` for your side's correlation and rate limiting. It's disclosed in the README and the docstring, which is why I'm fine with it — but it is opt-in telemetry to a third party, and that should be visible to anyone reading this PR.
- `all = [...]` now pulls `mcp>=1.8`, nudging the floor for `[all]` installers.

And one honest limitation for anyone who finds this later: there's no `fetch_many`, so bulk waves degrade to per-URL through the `hasattr` check at `cli/fetch_batch.py:116` and will be noticeably slower than crawl4ai on a wide sweep. You flagged that as a follow-up and I'm fine with it, but it means this is a good search provider rather than a drop-in replacement for the fetch path.

The thing I can't test is your endpoint. CI stubs the SDK, so the first signal that the schema changed will be a user bug report landing here. That's the cost I'm accepting by merging, and I'd rather say so than pretend otherwise.

jordan-gibbs:
Merged. Two things I did on the branch: moved the section into a new **Web providers** list where all five providers sit on equal footing, and swapped "free" for "keyless". One more that mattered: this PR had been red on CI since August 10 — `asyncio.run()` inside a running loop, the same class of bug #86 fixed. Added the loop-safe runner; first green run today. Thanks for the disclosure and for addressing the earlier notes properly.

PR #63, opened 2026-07-26
feat: add Codex platform integration

kaddynator:
## Summary

- add `hyperresearch install --platform claude|codex|both`, preserving Claude as the default
- install Codex instructions in `AGENTS.md` and skills under `.agents/skills/`
- translate Claude-only orchestration vocabulary into Codex-native skill routing
- package the 16 specialist definitions as reusable role prompts for generic Codex subagents
- keep Codex artifacts updated through profile changes, repair, and agent-doc refreshes

This supersedes #55, which was self-closed without review. In addition to the
installer adapter, this version covers profile/repair/doc-refresh lifecycle
behavior, preserves the existing Claude JSON contract, and includes explicit
Claude regression and mixed-platform tests.

## Backward compatibility

The default remains `claude`; no existing command needs to change.

| Command | Instruction file(s) | Skill tree(s) |
|---|---|---|
| `hyperresearch install` | `CLAUDE.md` only | `.claude/skills/` only |
| `hyperresearch install --platform codex` | `AGENTS.md` | `.agents/skills/` |
| `hyperresearch install --platform both` | both | both |

Regression coverage explicitly verifies:

- a default install does not create `AGENTS.md` or `.agents/`
- existing Claude golden prompts and hook tests remain unchanged
- the default `--steps-only --json` value keeps its original string-or-null shape
- an unrelated user-authored `AGENTS.md` does not enable or mutate the Codex integration
- changing profiles rerenders both platforms only when both were installed

## Codex implementation

- `.agents/skills` and `AGENTS.md` follow Codex's documented repo-scoped discovery paths
- the generated router uses explicit skill-file loading and the run manifest/Codex plan for recovery
- specialist prompts are progressive-disclosure references loaded before generic subagent delegation
- generated skills are tested to contain no Claude-only `Skill(...)`, `TodoWrite`, `Task` orchestration, `.claude/skills`, `subagent_type`, or Claude-in-Chrome references
- if subagent delegation is unavailable, the skills preserve a local-execution fallback

One platform distinction is documented rather than hidden: Claude Code supports
per-agent tool allowlists, while Codex subagents inherit the parent sandbox.
The Codex adapter preserves patcher/polish boundaries as explicit role
instructions; it does not claim an equivalent mechanical per-tool lock.

## Validation

- complete CLI/core/graph/search test suites
- complete web-provider test suite with optional provider dependencies
- Claude hook and golden-prompt regression suites
- Codex-only, both-platform, steps-only, global-install, idempotency, and profile-switch tests
- `ruff check src tests`
- strict mypy check for the new Codex and agent-doc modules
- wheel and source distribution builds
- live Codex desktop discovery: router + 18 step skills
- local install smoke test: 19 skills, 16 role prompts, successful vault status

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## [Unreleased]
 
+### Codex compatibility
+
+- **Additive platform installer:** `hyperresearch install --platform codex|both`
+  now installs Codex project instructions and skills while preserving the
+  existing Claude Code default and behavior.
+- **Native Codex layout:** the router and pipeline steps install under
+  `.agents/skills/`, with specialist agent definitions packaged as reusable
+  role-prompt references for generic Codex subagents.
+- **Durable Codex guidance:** `AGENTS.md` is created or updated idempotently
+  without overwriting user-authored content. Profile changes, repair, and
+  agent-doc refreshes keep whichever integrations are present up to date.
+
 ## [0.9.1] - 2026-07-25
 
 ### Four silent-failure leaks closed (tags, FTS syntax, batch PDFs, cache-busting date)
--- a/README.md
+++ b/README.md
@@ -67,6 +67,12 @@ Platform selection is additive and never removes another integration:
 | `hyperresearch install --platform codex` | `AGENTS.md` | `.agents/skills/` |
 | `hyperresearch install --platform both` | both files | both skill trees |
 
+The research workflow and artifact contracts are shared across platforms.
+Enforcement differs at one boundary: Claude Code's registered agents can use
+per-agent tool allowlists, while Codex subagents inherit the parent sandbox.
+The Codex role prompts therefore preserve the patcher/polish tool boundaries as
+explicit instructions rather than claiming an equivalent mechanical allowlist.
+
 ---
 
 ## The 16-step research pipeline
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 
 ---
 
-**Hyperresearch turns Claude Code into a deep research agent: one that currently leads the DeepResearch-Bench RACE leaderboard (benchmarked internally).** A tier-adaptive 16-step pipeline takes one prompt and produces an adversarially-audited report with full source provenance. Every source it reads lands in a persistent, searchable vault, so each session starts smarter than the last.
+**Hyperresearch turns Claude Code or Codex into a deep research agent: one that currently leads the DeepResearch-Bench RACE leaderboard (benchmarked internally).** A tier-adaptive 16-step pipeline takes one prompt and produces an adversarially-audited report with full source provenance. Every source it reads lands in a persistent, searchable vault, so each session starts smarter than the last.
 
 <p align="center">
   <img src="assets/benchmark.png" alt="DeepResearch-Bench top-5 hyperresearch leads the chart ahead of Grep Deep Research, Cellcog Max, nvidia-aiq, Gemini Deep Research, and OpenAI Deep Research" width="780">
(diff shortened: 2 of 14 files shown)


jordan-gibbs:
This has been open too long without a straight answer, so let me give you one: **I haven't decided whether this project supports a second platform, and that decision is the blocker — not the code.** It's a real question and I don't want to keep you guessing while I avoid it.

Separately from the direction question, there's one thing in here I'd need fixed regardless, and I want to flag it now because it's a regression of something I fixed after you opened this.

**`core/codex.py::install_codex_step_skills` deletes on a name match alone.** It walks `.agents/skills/` and `shutil.rmtree(child)` for any directory starting with `hyperresearch-` that isn't in the current roster. That's exactly #73 — someone lost a hand-written personal skill because the installer assumed a name prefix meant we owned the directory. #75 fixed it on the Claude path by adding an ownership check (`_is_our_skill_dir` at `core/hooks.py:3967`, gated at `:4165`), on the principle that a name match is not license to delete. The Codex path has no equivalent, and none of the seven tests in `tests/test_cli/test_install_codex.py` covers the prune. That one isn't negotiable whichever way the direction call goes.

The rest of my hesitation, so you can judge whether it's worth your time to keep pushing:

- **The adapter is string substitution over the prompt corpus.** `_codexify` does ~18 literal replaces plus two regexes. `content.replace("Claude Code", "Codex")` rewrites prose indiscriminately, and `"Claude-in-Chrome" → "authenticated browser integration"` turns the browser-escalation agent into a role prompt describing tools Codex doesn't have — so the documented escalation lane (blocked fetch → browser-fetcher) silently doesn't work there. The `subagent_type` regex also mangles the template at `hyperresearch-12-critics.md:39` into a path that will never exist.

- **The tool-lock invariant becomes advisory.** You're upfront that Codex subagents inherit the parent sandbox, so the patcher and polish-auditor Read+Edit lock survives only as prose. That invariant is load-bearing — it's what stops a patcher from regenerating a report instead of editing it.

- **It couples tightly to `hooks.py` internals** — five private functions and 16 agent constants pulled by `getattr` from a 4,178-line module. A rename anywhere in that set fails at runtime rather than import.

- **`CODEX_BLURB` has already drifted**, before the PR has even landed: #74 added 37 lines of open-access guidance to the Claude blurb on 2026-08-01 and the Codex one has none of it. That's the maintenance cost demonstrating itself.

Here's my honest position. For a project I maintain alone, the cost isn't the merge — it's that every future prompt or agent edit has to be validated in two renderings, one of which is produced by substitution I can't easily eyeball. That's the part I'm weighing, and it's not a reflection on the execution, which is careful.

I'll give you a yes or no rather than leaving this open indefinitely. If it's a yes, the asks are: the ownership check with a test, a re-render against post-#74 prompts, and a decision on what happens to the browser-escalation lane. If it's a no, I'll say so plainly and explain why, and I'm sorry for the wait either way.

Mechanically it's in decent shape — the conflicts are just `CHANGELOG.md` and the `pyproject.toml` version context line, both a minute's work. The staleness is semantic (it predates #74 and #75), not structural.

jordan-gibbs:
Here's the answer I said was coming: **yes**. I want a second platform, and the execution here is careful enough to build on. Three things before merge, in priority order:

1. **The ownership check.** `install_codex_step_skills` must not delete on a name match alone — mirror `_is_our_skill_dir` from `core/hooks.py:3967` and add a test that a user's own `hyperresearch-*` directory survives a prune. This is the #73 regression and it's the one I won't waive.
2. **Re-render against current prompts.** Main has moved a lot since July — #74's open-access guidance, and today a scholar discovery layer that rewrote the agent instructions. `CODEX_BLURB` needs to be regenerated from the current `HYPERRESEARCH_BLURB`, not hand-patched.
3. **A decision on the browser lane.** The string substitution turns the escalation agent into a description of tools Codex doesn't have. Either mark the lane unavailable on Codex explicitly (fine) or wire an equivalent — but not a prompt that describes something that can't happen.

The conflicts are just CHANGELOG and the pyproject version line; a merge from main will sort them. Sorry for the wait.

PR #64, opened 2026-07-27
Fix length-in-range verify check for CJK reports

tetra4rnav:
`verify_run`'s length-in-range check counted words via `report_text.split()`, which is meaningless for prose in scripts that don't delimit words with ASCII whitespace (Chinese, Japanese, Thai, Lao, Khmer, ...). A report squarely within its documented character-count target (see the step-10 skill's "words / chars (CJK)" column, e.g. argumentative: 20,000-25,000 chars) could register as a few hundred "words" against the English 5,000-10,000 word target and fail the ship gate outright, even though the report was well-formed and every other check passed.

Detection is a structural heuristic (`_lacks_word_boundaries`: average ASCII-whitespace-delimited token length), not a hardcoded script allowlist, so it generalizes to any non-space-delimited writing system without enumerating Unicode ranges -- and correctly leaves Korean alone, since unlike Chinese/Japanese, Korean orthography IS space-delimited.

The character-count targets applied when the heuristic fires are NOT hardcoded to CJK either: `char_targets_no_word_boundary` is a profile field, same as `word_targets`, overridable per deployment through the existing `.hyperresearch/config.toml` profile-overlay mechanism. The shipped defaults are calibrated for CJK because that's the only such script family this project has real usage data for -- a deployment serving Thai, Lao, Khmer, or another non-word-boundary script supplies its own calibration the same way it would override any other tunable, no code changes required.

Verified against the spec: `skills/hyperresearch-10-triple-draft.md`'s "words / chars (CJK)" table documents short=1500-6000, structured=6000-15000, argumentative=20000-25000 chars, identically for every profile (only the word-count half is profile-parameterized in that template). The shipped `char_targets_no_word_boundary` values match this table exactly, and premier/dissertation/light correctly inherit them unchanged (via `**_FULL`) since the table itself doesn't vary by profile either. The one gap: no profile's `dissertation` response_format has a documented CJK char target anywhere in the skills, so it correctly falls through to the generic `word_target * 3` estimate rather than a fabricated number.

Tests cover: the detector across Japanese, Chinese, and Thai (proving it isn't a CJK-specific allowlist) and Korean (proving it correctly does NOT trigger for a space-delimited script); a citation-heavy English report that must not false-positive; a CJK report passing on chars despite a near-zero word count; a regression guard that a CJK report far outside its char target still fails; and an end-to-end profile-override test proving a non-CJK script gets its own calibrated target via config, not the shipped CJK numbers.


taikin-snake:
Tested this branch against the Japanese vault that originally hit the gate in #71, as offered there. Environment: Windows 11 / Python 3.13 / the 0.9.1 uv-tool install with this branch's `src` on `PYTHONPATH`. The vault's run is a 61-source light-tier run with a Japanese `structured` report using wikilink citations.

**Ship gate on the real run**

| | `length-in-range` | full `run verify` |
|---|---|---|
| 0.9.1 baseline | FAIL — `626 words vs target 2000-5000` | `passed: false` |
| this branch | OK — `12083 chars vs target 6000-15000 (structured; char-based; ±20% tolerance)` | `passed: true` (9/9 checks) |

**Detector on real report files**

| Report | chars | split tokens | avg token chars | `_lacks_word_boundaries` |
|---|---|---|---|---|
| Japanese report from #71 | 12,083 | 626 | 18.24 | `True` → char branch |
| `example-reports/rl-exploration-trajectory-planning.md` | 86,652 | 11,209 | 6.71 | `False` → word branch |

The Japanese report clears the 15.0 threshold with real margin despite being citation-heavy (the wikilink keys account for most of its ASCII whitespace), and the repo's own English example report stays comfortably on the word branch.

**Config-overlay interaction** (the point raised in #71): my interim workaround profile — `[profile.light-ja]` with `extends = "light"` overriding `word_targets` with split-token equivalents — resolves cleanly under this branch and inherits the base profile's `char_targets_no_word_boundary` (`structured = (6000, 15000)`). The distorted word targets simply stop mattering for CJK reports, so the workaround profile can be deleted once this merges.

The shipped char targets also match the CJK ranges the step-10 skill documents (1500–6000 / 6000–15000 / 20000–25000), so the drafting end and the gate agree again. From the Japanese-vault side this looks ready.

PR #86, opened 2026-08-22
fix: run fetch_url off the running event loop, and drop the false DUPLICATE_URL on a re-fetched, deleted note

### LINKED ISSUE #84: Bug: fetch_url MCP tool fails — coroutine never awaited + false DUPLICATE_URL
### Environment
- hyperresearch 0.10.0 (installed via `pip install hyperresearch[mcp]`, so `mcp>=1.6,<2` per pyproject)
- Python 3.13.12 (venv)
- crawl4ai 0.9.2
- `hyperresearch mcp` stdio server

### Observed behaviour
The `fetch_url` MCP tool fails to create a note and emits a warning that the
crawl4ai fetch coroutine was never awaited. The plain CLI `hyperresearch fetch
<url>` works perfectly (fetched Wikipedia pages and produced 1.5k–7.7k-word
notes), so this is specific to the MCP `fetch_url` wrapper.

#### 1. "never awaited" on any URL
Calling `fetch_url` for a never-fetched URL (`https://example.com`) logs:
```
'Crawl4AIProvider._fetch_async' was
never awaited
```
and returns `ok: true` based on truthy content while **no note is created**
(`search_notes` returns total 0).

#### 2. False DUPLICATE_URL with `note 'None'`
Calling `fetch_url` for a real, never-fetched URL
(`https://en.wikipedia.org/wiki/Web_crawler`) returns:
```json
{"ok": false, "error": "URL already fetched as note 'None'", "error_code": "DUPLICATE_URL"}
```
The dedup check reports a note id of `None` and aborts, even when no such note
exists in the vault.

### Reproduction (via the official `mcp` client)
```python
from mcp import ClientSession, StdioServerParameters
from mcp.client.stdio import stdio_client
async def main():
    p = StdioServerParameters(command=".venv/bin/hyperresearch", args=["mcp"])
    async with stdio_client(p) as (r, w):
        async with ClientSession(r, w) as s:
            await s.initialize()
            print(await s.call_tool("fetch_url", {"url": "https://example.com"}))
asyncio.run(main())
```

### Expected
`fetch_url` should fetch the URL, store a note in the vault (like `hyperresearch
fetch`), and return the new note id — matching the documented behaviour that the
MCP server exposes the same vault as the CLI.

### Notes
The other 12 MCP tools (`create_note`, `search_notes`, `read_note`,
`read_many`, `get_backlinks`, `get_hubs`, `vault_status`, `lint_vault`,
`check_source`, `list_sources`, `list_notes`, `update_note`) all work
correctly. Only `fetch_url` is broken.

AmirF194:
## Root cause

Two independent bugs in the `fetch_url` MCP path, both in code the CLI's `hyperresearch fetch` never touches.

**1. `asyncio.run()` inside a thread that already has a running loop.** `Crawl4AIProvider.fetch()` calls `asyncio.run(self._fetch_async(url))`. The MCP stdio server dispatches synchronous tool functions directly on its own running event-loop thread (it does not offload them to a worker thread), so this call chain lands on a thread where a loop is already running. `asyncio.run()` in that case raises immediately, before the coroutine it was given ever executes, which is exactly the "coroutine ... was never awaited" warning plus a swallowed failure the issue describes. The CLI has no such loop and this path has always worked for it.

**2. False `DUPLICATE_URL` after a note is deleted.** `sources.note_id` is `ON DELETE SET NULL` (`db.py`), so deleting a note leaves an orphaned `sources` row with `note_id = NULL`. `fetch_and_save`'s dedup check only tested truthiness of the row (`if existing:`), so an orphaned row reads as a live duplicate forever, and the error message prints the id as the literal string `'None'`.

## Fix

- `crawl4ai_provider.py`: added `_run_coro()`, which checks for a running loop on the current thread. With no running loop it behaves exactly like before (`asyncio.run(coro)`); with one, it runs the coroutine to completion on a dedicated thread, which has no loop of its own. `fetch()`'s two call sites now go through it.
- `fetcher.py`: the dedup check now treats a `NULL` `note_id` as an orphan, not a duplicate, and the `INSERT` for a new source is an upsert (`ON CONFLICT(url) DO UPDATE ...`) so re-fetching a URL whose note was deleted updates the orphaned row instead of hitting the `url` primary key.

## Verification

- `tests/test_web/test_crawl4ai_running_loop.py`: new test drives `fetch()` from a thread with a real running event loop (via a helper thread, matching the pattern already used in `tests/test_web/test_fetch_many_fallback.py` for the same reason: a leaked event-loop-running flag from an earlier test in the full suite makes a bare `asyncio.run()` in the test itself unreliable). Fails on `main` with `RuntimeError: asyncio.run() cannot be called from a running event loop`; passes on this branch. A second test pins the CLI's no-running-loop path unchanged.
- `tests/test_core/test_fetcher_orphan_dedup.py`: new test deletes a fetched note, confirms the orphaned row (`note_id IS NULL`), then re-fetches the same URL and asserts it succeeds instead of raising. A second test confirms a real, live duplicate is still rejected.
- Ran the full suite (`pytest tests/ -v`, 703 passed) and `ruff check src/ tests/` in Docker on both Python 3.11 and 3.13 (this repo's CI matrix also runs 3.12); `python -m build` also verified.
- Not verified: the exact MCP-client-visible symptom (the "never awaited" warning plus `ok: true` with no note created) end to end through a live `mcp` stdio session, since that needs the real `mcp` package wired to a running server rather than a unit-level call into `Crawl4AIProvider.fetch()`. The underlying `RuntimeError` this fixes is the one the issue's traceback names.

Open PR #53 (SSRF gate) touches both files this PR changes but at different lines; read its diff and confirmed no overlap.

Fixes #84


AmirF194:
Checking in a week after opening this one, no rush at all. Happy to rebase or split it up if that would help move it along.

jordan-gibbs:
Both halves of this are real and I reproduced them.

The loop bug is exactly as you describe: the installed `mcp` dispatches sync tool functions inline on the running loop (no `to_thread` offload), and `fetch_url` is `def`, not `async def`, so `asyncio.run()` inside the provider raises and the coroutine is left unawaited. The MCP server's only write-side fetch tool has been dead.

The orphan dedup is the more interesting one, and you fixed the cause rather than the symptom. `sources.note_id` is `ON DELETE SET NULL` with foreign keys on, so `note rm` leaves the row behind with a null id, and the check at `core/fetcher.py:33` only tests row truthiness — which is why the message renders as `already fetched as note 'None'`. Since `url` is the primary key, that URL is then blocked permanently with no CLI path to clear it. The upsert is genuinely required; a plain insert would hit an integrity error.

CI is green. Two asks before merge:

**The same bug exists in three more places, all untouched here.** `core/fetcher.py::fetch_and_save` is called only by the MCP server — the "reusable by CLI and MCP" docstring at the top of that file is stale — and the CLI carries its own copy:

- `cli/fetch.py:246-247` → emits the identical `'None'` message at `:281`, and with `--suggested-by` passes `note_id=None` into `_append_suggested_by_to_existing`
- `cli/fetch_batch.py:82-83` → silently skips the URL
- `cli/research.py:112-116` and `:242-244` → same

Every `note rm` of a fetched note creates one of these, so the CLI path hits it as often as the MCP one. Factoring the check into a single helper would be ideal; fixing all four sites is fine too.

**`fetch_many` is still unwrapped** at `web/crawl4ai_provider.py:412`. Not on the MCP path today, but it's the same trap and your own rationale covers it.

One design note, take it or leave it: `_run_coro` joins the thread from the event-loop thread, so the server serves nothing for the duration of a fetch. Not a regression, and the provider-level fix is more defensive than fixing it at the boundary — but making `fetch_url` `async def` and awaiting `anyio.to_thread.run_sync` would keep the server responsive. Your call.

Last thing: I'm going to leave #84 open when this merges. Symptom 1 in that report — `ok: true` with no note created — isn't explained by this fix. `fetch_and_save` has no swallow, `get_provider` has no fallback, and the error path would have returned `FETCH_ERROR`. Your PR body is upfront that it wasn't verified end to end, so either the reporter conflated two runs or there's a third bug. Worth one more look at the original repro after this lands.

AmirF194:
Fixed all three CLI call sites (`cli/fetch.py`, `cli/fetch_batch.py`, `cli/research.py` x2) with the same `existing and existing["note_id"] is not None` check, and wrapped `fetch_many` in `_run_coro` too. Added a test file (`test_orphan_dedup_call_sites.py`) covering the orphan-refetch path through `fetch`, `fetch-batch`, and `_save_result` directly; confirmed all three fail on the pre-fix code and pass after.

On the design note: agreed `async def` + `anyio.to_thread.run_sync` is the better fix, but it touches the provider's fetch signature more than this PR's scope, so left it as `_run_coro` for now. Can open a follow-up if you'd rather not carry the responsiveness gap.

Rebased onto main, CI should confirm.

jordan-gibbs:
Merged — and this turned out to be bigger than either of us thought. Your later commit had already patched the three CLI sites, but the fix was inert there: `INSERT OR IGNORE` is a no-op on an orphaned row too, so nothing re-linked it. Worse, `hpr fetch`'s race detector read the still-NULL row as a lost race, **deleted the note it had just written**, and returned `ok: true, note_id: null`. That is the unexplained symptom 1 from #84. Added a `reclaim_orphaned_source_row` step after each insert, rewrote the tests to drive real `note rm`, and wrapped `fetch_many`. Good bug, well found.

jordan-gibbs:
Missed answering your question in the merge note: yes, please open the follow-up for the `async def` + `anyio.to_thread.run_sync` boundary. You're right that it was out of scope here, and the responsiveness gap is real — the MCP server serves nothing for the duration of a fetch. Shipped in v0.11.0 as-is; the follow-up can land whenever.

AmirF194:
Nice find on the reclaim step, that's a sharper fix than what I had. I'll open the follow-up for the async def / anyio.to_thread.run_sync boundary.

PR #90, opened 2026-08-25
Update CONTRIBUTING.md for the hyperresearch rename

MarceloSenai:
## What

`CONTRIBUTING.md` still described the project as **kasten** throughout. Every affected line is something a new contributor copies and runs:

- clone URL and `cd` target
- `mypy src/kasten/`
- `kasten --version` in the issue template
- the issues link
- `kasten sync --force` in the design-principle note

## Architecture overview had drifted further than the name

It listed three packages that no longer exist — `ingest/`, `compile/`, `llm/` — and omitted four that do: `web/`, `export/`, `mcp/`, `skills/`. Descriptions for the new entries are taken from each package's own module docstring rather than guessed.

## One judgment call worth flagging

The "Code style" section asked contributors to run:

```bash
ruff check src/
ruff format src/
```

`.github/workflows/` runs `ruff check src/ tests/` and no format step, and the tree isn't ruff-format-formatted — `ruff format --check .` currently reports 99 files would be reformatted. So a contributor following the doc produces a 99-file reformat that buries their actual diff. I aligned the doc with what CI enforces.

If the intent is actually to adopt `ruff format`, that's a separate (and much larger) change and this hunk should be dropped from the PR — happy to do that.

## Verification

`hyperresearch --version` and `hyperresearch sync --force` both checked against the installed CLI. Used the `hyperresearch` entry point rather than the `hpr` alias to match README, which uses it throughout. No `kasten` references remain in the file. Docs only — no code touched.


jordan-gibbs:
Accurate, and thanks for checking the `ruff` question against CI rather than guessing — you're right that both workflows run `ruff check src/ tests/` with no format step, so dropping `ruff format src/` from the docs is correct and avoids sending contributors into a repo-wide reformat.

The package list matches the tree now, and I confirmed the three you removed (`ingest/`, `compile/`, `llm/`) are genuinely gone.

Two lines in the same file before I merge, both adjacent to what you're already touching:

- **Line 47** still says ``Run `ruff check src/` and `python -m pytest tests/` `` — the old narrower invocation. You fixed the identical command at line 28, so as it stands the file now contradicts itself.
- **Line 21** claims "The test suite runs in under 30 seconds." It's ~700 tests now and nowhere near that. Not a rename issue, but it's two lines up and it's the kind of thing a new contributor trusts and then worries about.

For what it's worth, `CONTRIBUTING.md` is the last file with live stale `kasten` references — the only other hits in the repo are the fork note in `CHANGELOG.md`, which is correct history, and "Zettelkasten" in `bench/`, which is a different word. So this finishes the rename.

jordan-gibbs:
Merged, after fixing the two lines myself (line 47's ruff invocation and the stale 30-second claim). Thanks — CONTRIBUTING was the last file with live `kasten` references.

PR #95, opened 2026-08-30
fix: verify keys tier artifacts on the classified tier, not the manifest profile

### LINKED ISSUE #68: run finish blocks every light-tier run: verify keys tier artifacts on the manifest profile, not the classified tier
The router's bootstrap says to initialize the run with the installed gear ("For a standard run, pass the installed gear (`full`)") and notes that "if step 1 classifies differently, the manifest's profile field is informational; the decomposition's tier rules."

`verify_run` does not honor that: it derives the required artifact set from `manifest["profile_steps"]`. A run initialized with `--profile full` whose decomposition then says `"pipeline_tier": "light"` is required to produce `critic-findings-*.json` and `patch-log.json`, which light tier correctly never creates (steps 12 and 14 are skipped by the tier gate). The run finishes all its light-tier work and then sits at `blocked (verify)` with no legitimate way to pass, since the gate's own rules forbid re-interpreting failures and stubbing fake critic findings would defeat the point of the gate.

Reproduction:

1. `hpr run init t1 --profile full`
2. Write `research/runs/t1/prompt-decomposition.json` with `"pipeline_tier": "light"` plus a valid light run (report with required headings, polish-log.json).
3. `hpr run finish t1` fails on `artifact:critic-findings-dialectic.json` etc.

Suggested fix (what we ship in our fork): in `verify_run`, when the decomposition declares a `pipeline_tier` that differs from the manifest profile, resolve that tier's profile and take the step set from it, falling back to `profile_steps` when the tier is absent or unknown. One regression test: a light-classified run initialized with the gear profile passes with only `polish-log.json` required. 

Happy to send the PR.

maximilliangrand:
Fixes #68.

## Root cause

`verify_run` derived its required-artifact set from `manifest["profile_steps"]` (`src/hyperresearch/core/runs.py:489` on main). The router documents the opposite rule: initialize the run with the installed gear, and "if step 1 classifies differently, the manifest's profile field is informational, the decomposition's tier rules" (`src/hyperresearch/skills/hyperresearch.md`, bootstrap step 2.5).

So a run started with `--profile full` whose `prompt-decomposition.json` says `"pipeline_tier": "light"` was still gated on steps 12 and 14, which the light tier skips. `run finish` demanded `critic-findings-*.json` and `patch-log.json`, the run did all of its light-tier work, and then sat at `blocked (verify)` with no legitimate way to pass.

Diagnosis and the fix shape are @fduple's, from the issue.

## Fix

`_required_step_ids()` in `src/hyperresearch/core/runs.py` resolves the tier declared in the decomposition when it disagrees with the manifest profile, and takes the step set from that profile. A decomposition with no tier, an unreadable one, or a tier that resolves to no profile all fall back to `profile_steps`, so gear runs and pre-tier runs keep the artifact set they have today. `full`, `premier` and `dissertation` resolve to the same step tuple, so only light-classified runs change.

The manifest profile still supplies every other verify input (word targets, char targets): the gear owns the numbers, the tier owns the routing.

## Tests

`TestClassifiedTierArtifacts` in `tests/test_core/test_verification.py`:

* a light-classified gear run passes with only `polish-log.json` present
* a gear run with no declared tier still fails on the missing critic findings
* an unknown declared tier falls back to the manifest profile

The first test fails on main and passes with the fix. End to end, the issue's repro goes from `BLOCKED (verify)` to `SHIPPED`. Full suite: 701 passed. `ruff check src/ tests/` clean.

Assisted by Claude (Opus), disclosed in the commit trailer.


jordan-gibbs:
The fix is right and I want it. I reproduced the bug, and your read of it matches mine: `core/runs.py:489` keys the artifact requirement on `profile_steps`, which is stamped once at init from the gear profile and never revised, so a run started on `full` and classified `light` is gated on step 12 and 14 artifacts the tier correctly never produces. That contradicts the router's own rule at `skills/hyperresearch.md:106` — the manifest profile is informational, the decomposition's tier is authoritative.

`_declared_tier()` + `_required_step_ids()` is the right shape, the fallbacks are correct, and `test_light_classified_gear_run_passes_without_critic_artifacts` fails cleanly without the change. CI is green on all three versions.

The problem is the diff. It's 118KB for a ~42-line fix, and the rest is an em-dash strip: **159 lines removed containing an em-dash, zero added.** 146 of those are `CHANGELOG.md`, i.e. the project's entire release history rewritten. Could you drop all of it and resubmit just the fix?

Specifically, beyond the CHANGELOG:

- `core/runs.py:481` changes a **user-facing string** — `f"{len(errors)} error(s) — first: "` → `f"{len(errors)} error(s), first: "`. That changes what `hpr run verify` prints, unrelated to this bug.
- The `runs.py` module docstring, the `add_spend` comment, and the `verify_run` / `finish_run` docstrings.
- `tests/test_core/test_verification.py:197`, a wire-cluster fixture string for a different feature.

The substitutions also aren't neutral where they landed in prose — "skips Unpaywall entirely — their terms require a real address" became a comma splice, and a few others turned parenthetical asides into run-ons.

I'm assuming this was an automated pass that picked up more than intended rather than a deliberate style change; your #96 is clean, which is why I read it that way. If you'd rather propose a CHANGELOG style change separately I'm open to hearing it, just not bundled with a bug fix.

Two small things on the fix itself while you're in there:

- `_declared_tier` catches `json.JSONDecodeError` but not `OSError`, so an unreadable-but-present decomposition file propagates where the docstring promises a fallback.
- Not a blocker, and not your scope: the same `profile_steps` assumption survives in `run resume` / `run status` at `core/runs.py:249-252`, so a light-classified run still gets told its next step is 3. I've filed that separately as #100.

jordan-gibbs:
Merged, with the diff reduced to the fix. I took main's CHANGELOG, kept `_declared_tier` / `_required_step_ids` / the call site / your three tests, restored the em-dashes in the docstrings and the `run verify` output string, and added the `OSError` fallback. Final diff: 102 insertions, 2 deletions, zero em-dash removals. The fix itself was right the first time.

PR #97, opened 2026-09-11
fix: send the PreToolUse reminder as hookSpecificOutput JSON (#94)

### LINKED ISSUE #94: PreToolUse hook writes only to stderr with exit 0, so its reminder never reaches the model
## Summary

The `PreToolUse` hook installed by `hyperresearch install` writes its reminder to `stderr` and exits 0. Under Claude Code's hook contract, a hook's `stderr` is surfaced to the model only on exit code 2. On exit 0 it goes to the debug log. So the reminder never reaches the agent, and the hook does not do the one thing it exists to do.

Version: `hyperresearch` 0.10.0 (PyPI).

## Evidence

`.hyperresearch/hook.js` has a single output statement, at line 38, and no `process.stdout` write and no explicit exit:

```js
const vault = findVault();
if (vault) {
    const msg = [
        'HYPERRESEARCH: A research knowledge base exists in this project.',
        // ... reminder text ...
    ].join('\n');
    process.stderr.write(msg + '\n');
}
```

Measured against the session transcript, over a fixed sequence of 5 `Grep` plus 5 `Glob` calls:

- The hook fired 10 times.
- The payload text appears in 10 transcript records, and all 10 are `hook_success` log entries with `exitCode: 0`, empty `content` and empty `stdout`. The text lives only in `stderr`.
- The payload text appears **zero** times in any `user`-type record, which is the channel that carries tool results and injected context.
- The non-cached `input_tokens` value is constant across the turns, leaving no room for a side channel.

A control run of the identical sequence with the hook removed produced 0 payload occurrences and 0 `PreToolUse` firings, against 10 and 10 with it installed. So the firing is real; the delivery is not.

Measured cost while delivering nothing: roughly 34 ms per call, most of it Node process startup, on every `Glob`, `Grep`, `WebSearch` and `WebFetch`.

## Why it matters

The hook's stated purpose is to remind the agent to search the existing research base before going to the web, and to prefer `hyperresearch fetch` over `WebFetch`. That advice is good, and users who install the tool are entitled to assume it is being delivered. Today the agent never sees it, so the behaviour the hook is meant to produce depends entirely on the separate `CLAUDE.md` block.

The failure is also silent in both directions: there is no error, and the hook reports success.

## Suggested direction

Emit the reminder as JSON on `stdout` using the documented injection field rather than writing to `stderr`:

```js
process.stdout.write(JSON.stringify({
  hookSpecificOutput: {
    hookEventName: 'PreToolUse',
    additionalContext: msg
  }
}));
```

That is the supported path for putting text in front of the model from a `PreToolUse` hook on exit 0. We have four hooks on this machine using that shape, including one whose job is almost identical to this one, and the text does reach the model.

If the intent was deliberately to log rather than to inject, then the fix is the opposite one: say so in the README, because the current wording reads as though the agent is being reminded.

## Note on frequency

Whichever direction is chosen, the matcher is worth revisiting. The same static text on every `Glob` and `Grep` call is a lot of repetition for a reminder that does not change. If it is injected as written, it would be re-sent on every matching call with no deduplication, which is a real context cost on long sessions. A once-per-session injection would deliver the same advice at a fraction of the price.

dajiaohuang:
Fixes #94.

`hyperresearch install` writes `.hyperresearch/hook.js`, and that script printed its reminder with `process.stderr.write(...)` and then exited 0. Claude Code only surfaces a hook's stderr to the model when the hook exits 2. On exit 0 stderr goes to the debug log, so the reminder never reached the agent, and the hook did not do the thing it exists to do.

The script now emits the documented injection payload on stdout instead:

```js
process.stdout.write(JSON.stringify({
    hookSpecificOutput: {
        hookEventName: 'PreToolUse',
        additionalContext: msg
    }
}) + '\n');
```

Everything else about the hook is unchanged. It is still registered on the same `Glob|Grep|WebSearch|WebFetch` matcher, it still exits 0, and the vault check that gates the message is untouched, so the reminder still only appears when the project actually has a vault. stderr is now empty.

## Verification

Rendered the installed script the way `install` does and ran it under node in a project with a vault, before and after:

```
before: exit 0, stdout '', stderr 570 bytes
after:  exit 0, stdout {"hookSpecificOutput":{"hookEventName":"PreToolUse","additionalContext":"..."}},
                 parses as JSON, stderr 0 bytes
```

Two tests cover both halves. One asserts the rendered script carries the injection contract and no longer writes the reminder to stderr. The other runs the installed script under node and reads the channel the model reads, checking the exit code, the empty stderr, and the parseable payload. The node test skips when node is not on PATH, so the suite still runs where it is not installed.

700 tests pass, ruff clean.


jordan-gibbs:
You're right about the contract, and I verified it independently rather than assuming: for PreToolUse at exit 0, neither stdout nor stderr reaches the model — plain stdout only reaches it for `UserPromptSubmit`, `SessionStart` and `PostModelSwitch` — and `hookSpecificOutput.additionalContext` with `hookEventName` is the documented channel. So the reminder this repo has shipped since it was introduced has never once been delivered. The payload change is correct, the tests are the right tests, and the node-subprocess one actually parses the output rather than asserting on a string.

Also worth noting in favour of merging: `_write_hook_script()` runs at `core/hooks.py:3704`, *before* the already-installed early return, so this reaches existing installs on a re-run of `install` rather than only new ones.

One thing I'd like in the same PR before I merge. Right now the hook costs nothing because its output is discarded. After this, it injects ~570 bytes of identical static text on every `Glob`, `Grep`, `WebSearch` and `WebFetch` — and I don't think cross-turn dedup saves us. The documented dedup is for identical strings from *multiple hooks* in one dispatch, which is a different thing.

Your own issue raised this in its last paragraph and the PR doesn't engage with it, so: could you narrow the matcher to `WebSearch|WebFetch`? On `Glob` and `Grep` the advice is noise — those are local vault operations, and "check the research base before searching the web" doesn't apply. That keeps the injection on the two tools where it's actionable.

If you'd rather keep the wider matcher, the alternative is the `once` field, but I'd want to know what it does to a persisted `settings.json` before relying on it — the description says the hook is *removed* after execution, which sounds like it mutates the file. Narrowing the matcher is the boring option and I'd take it.

Unrelated heads-up: #98 touches the same `[Unreleased]` anchor and the same import block in `tests/test_core/test_hooks.py`. It's merged now, so you'll want to rebase — it'll be a trivial conflict.

jordan-gibbs:
Merged. I narrowed the matcher to `WebSearch|WebFetch` myself with a comment explaining why, and resolved the trivial conflict with #98. Your contract analysis was correct in every particular — confirmed against Claude Code's embedded hook docs.

Issue #101, opened 2026-09-11
Pipeline constants are authored twice: profile fields vs hardcoded values


jordan-gibbs:
Split out of #88 (contract audit by @darlingm). This is the root cause behind a large share of that audit, and it's worth fixing as a class rather than case by case.

The pattern: a tunable is declared on the profile **and** hardcoded somewhere that actually runs. The profile field then drifts, or is dead on arrival, and two halves of the same run disagree.

### Confirmed instances

**`citation_density_min` is dead config.** Declared at `core/profiles.py:140`, set to `2.0` at `:229`, and read **nowhere** — grep returns only the declaration and the assignment. Both real gates hardcode a different number: `core/runs.py:413` (`floor = 1.5`) and `core/hooks.py:1220`. This overlaps #76, which should be fixed in the same pass.

**Draft orchestrator and synthesizer get different word targets.** `core/hooks.py:2030-2032` hardcodes `argumentative: 5000-10000 words`, while `core/hooks.py:2261-2263` templates `<< p.word_targets[...] >>`. On the `premier` gear (`core/profiles.py:295` = 8000–16000) the two halves of the same run are handed targets 60% apart.

**`char_targets_no_word_boundary` is defined only on `_FULL`** (`core/profiles.py:238`). `premier` and `dissertation` build from `**_FULL` while overriding `word_targets`, so they inherit full-gear character targets that no longer correspond to their own word targets. Same bug class, introduced recently in #64.

**Step 4 hardcodes what it templates.** `skills/hyperresearch-4-loci-analysis.md:33` templates `<< p.loci_analysts >>` but the same line says "ONE message, **both** Task calls", `:42` hardcodes `analyst_id: "a"/"b"`, `:44` hardcodes `loci-a.json`/`loci-b.json`, and `:51` says "Wait for **both**". All of it breaks on `premier`, where `loci_analysts: 3` (`core/profiles.py:278`). Likewise `:55` templates `<< p.loci_max >>` while `:59` hardcodes "exceeds **6**" (premier is 10, dissertation 20).

### Suggested fix

For each: template it, or delete the profile field. The failure mode is having both.

A cheap guard against regression: a test that asserts every `Profile` field is referenced at least once outside `profiles.py`, which would have caught `citation_density_min` at authoring time.

PR #115, opened 2026-09-11
fix(runs): reconcile spend from disk and record pid/heartbeat

### LINKED ISSUE #92: Spend counters stay at zero on a completed run, so --budget can never fire
## Summary

On a run that finished with `status: "done"`, every counter in `manifest["spend"]` is still zero, while the vault on disk holds 153 notes and 56 fetched PDFs. Because `--budget` is evaluated against `spend["estimated_usd"]`, the ceiling can never fire.

A second symptom of the same subsystem: the manifest carries no heartbeat and no PID, so a run whose host session dies stays `"status": "running"` indefinitely. There is no way to ask the manifest whether the run is still alive.

Version: `hyperresearch` 0.10.0 (PyPI), Python 3.13.

## Evidence

Run started with `--budget 25`. `run.json` after the run reached `done`:

```json
"budget_usd": 25.0,
"status": "done",
"spend": {
  "estimated_usd": 0.0,
  "sources_fetched": 0,
  "notes_written": 0,
  "agents_spawned": 0
}
```

On disk for that same run: 153 note files and 56 PDFs.

The counters are not reconciled at the end either. This is the final state, not a mid-run snapshot.

## Why the counters stay at zero

`core/runs.py` has exactly one function that increments spend, and it takes the amounts as explicit arguments:

```python
# core/runs.py:171-179
def record_spend(..., estimated_usd: float = 0.0, sources_fetched: int = 0, ...):
    spend["estimated_usd"] = round(spend.get("estimated_usd", 0.0) + estimated_usd, 4)
    spend["sources_fetched"] = spend.get("sources_fetched", 0) + sources_fetched
```

The budget gate reads that value:

```python
# core/runs.py:187
if budget is not None and spend["estimated_usd"] >= budget and manifest["status"] == "running":
```

But nothing on the pipeline path calls it. Grepping the 19 installed skills for `run spend` returns **0 matches**. The fetching and note-writing steps never report what they spent, so the accumulator stays at its initial `0.0` and the comparison at line 187 is never true.

The `--budget` help text is careful to say this is "a value measure, not a bill, on subscription billing", which is fair. The problem is not the unit. It is that the measure is never taken, so a user who sets a ceiling gets no protection of any kind, and no warning that the ceiling is inert.

## Second symptom: the run is not self-diagnosable

Top-level keys of `run.json`:

```
blocked_on, budget_usd, chapters, levers, manifest_version, profile,
profile_steps, spend, started_at, status, steps, updated_at, vault_tag, verify
```

There is no `pid` and no heartbeat timestamp distinct from `updated_at`. When the host agent session is interrupted mid-run, the manifest keeps `"status": "running"` forever. Anyone consulting the manifest concludes the run is alive. The only way to find out otherwise is to look at the operating system process table, which is outside anything the tool exposes.

These two belong together: both are the run failing to report its own reality.

## Impact

A user who runs a `full` tier job and sets `--budget` reasonably believes there is a stop. There is not. In our case the actual host-session cost of one completed `full` run was roughly 8.6x the ceiling that had been declared, and nothing in the tool's own output would have revealed it. Combined with the manifest problem, a run that has died silently also looks like a run that is still working, so the natural reaction is to wait rather than to check.

## Reproduce

1. `hpr run init --budget 1` on a fresh vault.
2. Run any `full` tier job to completion.
3. `jq '{budget_usd, status, spend}' research/runs/<tag>/run.json`.

Expected: non-zero `sources_fetched` and `notes_written` matching the files written, and `estimated_usd` above the ceiling triggering the block.
Actual: all four counters at `0`, `status: "done"`, and the notes and PDFs present on disk.

For the second symptom: start a `full` run, kill the host session mid-run, then read `run.json`. It reports `running` with no way to tell it is not.

## Suggested direction

Have the fetch and note-writing steps call the existing `record_spend` with real quantities, so the accumulator reflects work actually done. Separately, write a heartbeat timestamp (or a PID) into the manifest and treat a stale heartbeat as `interrupted` rather than `running`, so `run status` can tell the two apart.

Zhao0335:
﻿## What this changes

Reworked per review. The gate lives in **`set_step`** — every skill already
calls `hpr run step --status done`, so no prompt changes are required.

- **Counts** are reconciled from the vault DB by `vault_tag` (`tags.tag`) and
  each note's `raw_file` frontmatter — not mtime, which concurrent runs and
  `note update` / `repair` / `sources score` rewrite.
- **Ceilings** are observable CLI knobs: `hpr run init --max-sources` /
  `--max-notes`, checked at step-done. No `DEFAULT_USD_PER_*` unit prices.
  Agent-reported `estimated_usd` is unchanged.
- **`pid` / `heartbeat_at` are gone.** Stall detection uses `updated_at`
  (`possibly_stalled`); a CLI subprocess PID is already dead by the time
  anyone looks.
- **Resume:** `run resume` unblocks; `run step --status done --force` clears
  a *count* block so the next step-done can proceed when no new files
  landed. Force cannot clear `budget` / `verify`. Force-clears are capped
  (`counts_force_cleared`, max 3).

Fixes #92

## Why

Pipeline agents almost never call `hpr run spend`, so a completed run
reported all-zero counters while the vault held hundreds of notes and PDFs.
`--budget` only watches `estimated_usd`, which stays 0 without unit prices
this repo does not carry. The first cut of this PR invented those prices and
recorded a CLI `pid`/`heartbeat` — both rejected in review.

## How it was verified

- `pytest tests/test_core/test_runs.py` — pass
- `ruff check` on touched files — pass
- New tests: tag-filtered counts (untagged notes excluded), max_notes /
  max_sources block on step-done, blocked done requires force, force does
  not clear budget, force-clear progresses resume, force-clear cap

---

**Scope check**

- [x] No unrelated reformatting outside the lines this change needs
- [x] `CHANGELOG.md` under `[Unreleased]` only
- [x] No edits to unrelated fixtures


jordan-gibbs:
Thanks for taking this on. The `set_step` hook is the right place for it: every skill already calls `run step --status done`, so a gate there needs no prompt changes. The Windows `_pid_alive` is done carefully too. I can't merge this shape though.

The ceiling only fires because of `DEFAULT_USD_PER_*` at `runs.py:60-62`. Those are pricing numbers, which this repo doesn't carry, and on the run in #92 they come out to 2.65 against a ceiling of 25, so the exact reproduction still finishes `done`. Drop the coefficients and have the gate watch what the CLI can actually see: an optional `--max-sources` / `--max-notes` on `run init`, checked against reconciled counts at step boundaries.

For the counts, don't use mtime. `note update`, `repair`, and `sources score` all rewrite old notes, and concurrent runs share the vault, so the number drifts with time instead of tracking the run. Every note the pipeline fetches is already tagged with the `vault_tag`. Count by tag with a DB query, and get raw files through those notes' `raw_file:` field.

Drop `pid` and `heartbeat_at`. `os.getpid()` in `run init` is the CLI subprocess, which has exited by the time anyone looks, so `run status` says "host session likely died" on every healthy run. I reproduced that. `heartbeat_at` is written from the same `now` as `updated_at`, so it adds nothing, and `possibly_stalled` already covers liveness.

Two smaller things: mention the `set_step` hook in the PR body, since it changes every step boundary and the description doesn't say so, and check that `run resume` can get past a block. Right now the next step-done re-blocks even with no new files.

Zhao0335:
Reworked in 86d9cda per your review:

- Gate is set_step (every hpr run step --status done); no prompt changes.
- --max-sources / --max-notes on 
un init; no DEFAULT_USD_PER_*.
- Counts by ault_tag in the DB + 
aw_file, not mtime.
- Dropped pid / heartbeat_at; possibly_stalled uses updated_at.
- 
un step --status done --force clears a *count* block only (not budget/verify) so resume can proceed when no new files landed.

Ready for another look.

jordan-gibbs:
Thanks for the rework, this is the shape I was hoping for. I ran it against a run that just finished here (91 notes, 87 fetched pages) and it reports `notes_written: 91` and `sources_fetched: 0`. The count only sees notes with a `raw_file`, which means PDFs, and most of what a run fetches is HTML. Every fetch already lands a row in the `sources` table (`cli/fetch.py` writes it with the `note_id`), so joining that to `tags` on the vault tag gives the number the counter is named for:

```sql
SELECT COUNT(DISTINCT s.url) FROM sources s JOIN tags t ON t.note_id = s.note_id WHERE t.tag = ?
```

That query returns 87 on the same run.

One more thing while you're in there: `_count_agent_spawns` looks for `agent_spawn` / `spawn` / `agent` events and no skill emits any of those, so it always returns the manifest value. I'd drop it rather than leave a counter that can't move.

With the sources count fixed I'll merge.

jordan-gibbs:
The Sep 14 push is just a merge of main, so the two things from my last comment are still open: `sources_fetched` still counts only notes with a `raw_file` (a vault with 5 HTML notes and 1 PDF reports 1), and `_count_agent_spawns` is still called.

A few more came up reading it again:

- After `run resume` or `--force`, the next step marked done blocks the run again on unchanged counts. A forced clear should hold until the count actually goes up.
- Once the force cap is hit, `--force` still records steps as done while the run stays blocked, so the cap doesn't stop anything.
- `run reconcile` catches only `RunError`, so a bad tag gives a traceback. The rest of `run_cmd` catches `VaultError` too.
- The changelog changes the en-dashes in the #101 entry, and the new heading sits above existing bullets.
- The PR title still says pid/heartbeat.

Please also say plainly in the description that `--budget` itself stays inert, since nothing calls `run spend`. The count ceilings only protect real runs if the entry skill passes them to `run init`.

Zhao0335:
Addressed the open review points in `9582456` and merged current `main` in `71a3507` (the PR is mergeable again).

- `sources_fetched` now counts distinct `sources.url` values joined to the run's tagged notes. The regression case has five HTML notes and one PDF and reports six sources.
- Removed `_count_agent_spawns`; reconciliation preserves only explicitly reported `agents_spawned`.
- Count-block clears from either `run resume` or step `--force` save the current count, so later steps with unchanged counts proceed. A higher count blocks again. The three-clear cap is checked before writing the next step, and `run resume` cannot bypass it.
- `run reconcile` catches `VaultError`, including a bad tag, and returns a normal CLI error.
- The changelog entry is an Unreleased bullet after the existing entries, and the #101 en-dashes are restored. The PR title and description are updated.

The description states the remaining limits plainly: `--budget` stays inert in normal pipeline runs because nothing calls `run spend`; count ceilings affect real runs only if the entry skill passes them to `run init`.

The CLI and run tests pass after the merge (one optional crawl4ai test skipped), as do ruff and `git diff --check`. GitHub CI is queued.

Issue #31, opened 2026-06-08
Capture bibliographic metadata (journal, DOI, year, preprint-vs-published) for paper notes


MKrbm:
## Problem
When a paper is fetched, the note only stores the raw `source:` URL, and
`_detect_content_type` lumps everything into `content_type: paper`. So journal,
volume, etc..  are never captured but arXiv preprint
is indistinguishable from a peer-reviewed journal article. That distinction is
useful info we currently throw away.

## Motivation
- The vault compounds over time; it's far more valuable if each paper carries
  real bibliographic identity, not just a URL.
- Preprint vs published, venue, etc.. are worth keeping on the note. 

## Proposed strategy
A standalone, opt-in `hr biblio` command (working name). Does not touch the
fetch path or the 16-step pipeline.

- Scans `content_type: paper` notes, parses the DOI/arXiv ID from the existing
  `source:` URL, queries OpenAlex (no key), writes bib fields into frontmatter.
- Confident-ID match only (no title-search → no mis-linking). fail-open on any
  error. Fills empty fields by default; `--force` to refresh (can update preprint→published).
- New optional `NoteMeta` fields: `doi, journal, volume, issue, pages, year,
  authors, openalex_id, pub_type, bib_enriched_at` (+ a preprint/venue field).
- Add matching SQLite columns via idempotent migration; `sync` rebuilds them from
  frontmatter. Surface the fields in `note show` / `search` / MCP read tools.

## Out of scope (for now)
No BibTeX export, no change to report `## Sources`, no fetch-time auto-enrichment.

MKrbm:
If you are comfortable with this, I can create a PR for it.

jordan-gibbs:
Yes, send the PR! That's pretty much the shape that I'd have asked for: a standalone opt-in command that only matches confident IDs and fails open. Two requests going in:

1. Keep the OpenAlex client on plain httpx (it's already a base dep) so `hpr biblio` doesn't grow the install.
2. Add a `--dry-run` flag that prints what it would write. It mutates frontmatter across a lot of notes and I want a preview mode before anyone points it at a big vault.

`bib_enriched_at` for staleness is a good call, keep it.

jordan-gibbs:
[...]

PR #119, opened 2026-09-11
feat(search): select sources by scholarly metadata

maximilliangrand:
## What this changes

Adds DOI, venue, minimum citation count, retraction status, and retrieved OA version filters to CLI and MCP search. Users can browse metadata without a text query, or combine these filters with keyword and semantic search; results include the matching metadata.

Refs #31, specifically the queryability gap raised in the maintainer's follow-up. This builds on the existing metadata pipeline and leaves MKrbm's proposed bibliographic enrichment work separate.

## Why

A research vault can already store this metadata, but selecting sources or finding unchecked evidence currently requires manual inspection. For example:

```sh
hpr search "retrieval" --retraction not-retracted --min-citations 20 -j
hpr search --retraction unchecked -j
hpr search --oa-version submittedVersion -j
```

`not-retracted` matches explicit false only. The v13 cache migration restores the missing unknown state: old zeroes become NULL until normal sync reads their actual frontmatter state. The upgrade preserves related data, invalidates only cache fingerprints, and never rewrites Markdown.

The Semantic Scholar integration does not fetch retraction data, so arXiv and DOI fallback scoring preserve unknown through metadata, SQLite, and frontmatter. Existing notes scored with the old inferred false can be refreshed with `hpr sources score --fresh`; the cache migration alone cannot correct source Markdown. OA version describes the retrieved text, not a certification of peer review.

## How it was verified

- Full suite: **1,305 tests passed** after merging current main.
- Offline coverage includes CLI/MCP metadata browsing, combined filters, JSON null/false/true, pagination, invalid values, semantic exclusion before top-k, migration preservation, rollback, and incremental resync.
- Four new Semantic Scholar regressions fail on the pre-fix source. Both arXiv and DOI fallback preserve unknown on fresh scoring of previously enriched notes and after forced resync; unknown notes match `unchecked` and are excluded from `not-retracted`.
- `ruff check src/ tests/` and `git diff --check` pass.
- `mypy src/hyperresearch/` is blocked by the same eight dependency/type-stub errors reproduced on current main and this branch, including NumPy stubs using syntax newer than the configured Python target.

Documentation: `docs/source-search.md`. No new dependencies or external requests for filtering.

---

- [x] No formatter or automated cleanup pass applied


jordan-gibbs:
I like this a lot, and it's the piece of #31 I said mattered most. I ran the v13 migration against a real vault here (93 notes, tags, claims, embeddings): rebuilt cleanly, indexes back, foreign keys intact, and the filters behave.

One thing I'd ask before merging, because it decides whether `not-retracted` means what the docs say. `core/scholar.py:303` and `:337` return `is_retracted: False` from the Semantic Scholar paths, and S2 carries no retraction data at all. So a note scored through S2 (every arXiv id, and any DOI OpenAlex missed) gets an explicit false written to frontmatter and the DB without anyone having checked, and with this PR it then matches `--retraction not-retracted`, which the docs describe as explicitly checked. Returning `None` from those two paths and letting `scholar.py:452-472` write null through would make the filter honest. It's the same distinction your migration goes to the trouble of restoring.

Not asking for the remaining #31 fields here; those stay with MKrbm.

maximilliangrand:
Both Semantic Scholar paths now return `None`, and scoring preserves that unknown state through SQLite and frontmatter. Added regressions for arXiv and DOI fallback that verify the note remains `unchecked`, never matches `not-retracted`, and stays unknown after a forced sync. The tests also cover refreshing an already-scored note that had the old inferred `false`; the docs call out `hpr sources score --fresh` for those existing notes.

Merged current main and resolved the changelog conflict without changing the other #31 fields. All four new regressions fail before the fix; the full suite passes with **1,305 tests**, and ruff passes. Mypy still stops on the same eight dependency/stub errors reproduced on current main.

jordan-gibbs:
Checked the retraction chain end to end. Unknown stays null from lookup through scoring, frontmatter and sync, and the four regressions fail when I put the old `False` back. Ran v13 on a copy of an 804-note vault: counts, indexes and FK checks all clean. Merging.

One thing this makes visible that isn't yours: if OpenAlex errors during a `--fresh` sweep, the S2 fallback now writes null over a stored retraction (main wrote 0). I'll fix that separately so a known retraction is never overwritten by an unknown.

PR #123, opened 2026-09-12
feat: oracle seats (optional decomposer + chief-editor agents on a stronger model)

jordan-gibbs:
## Summary

Adds two optional "oracle seats" to `ModelMap`, both off by default, so a run that keeps its token-heavy work on a cheap fast model can put a top model in the single-call positions where judgment steers everything downstream:

- **`decomposer`** (default `"orchestrator"`): when set, step 1 spawns a new `hyperresearch-decomposer` agent on that model to produce the decomposition, coverage matrix and tier rationale. The agent executes procedure items 1–9 from the installed step-1 skill file (single source of truth); the orchestrator still renders levers and audits the artifacts against the exit criteria.
- **`chief_editor`** (default `"off"`): when set, step 12 spawns one `hyperresearch-chief-editor` on that model in place of the four critics. It reads the report once through all four lenses and writes the same four `critic-findings-*.json` files, so gap-fetch, the patcher and `run verify` are unchanged.

With the existing `corpus_critic` seat, that covers the three positions with the best judgment-per-token ratio: pre-research (what is being asked), pre-draft (what would overturn the corpus), and post-synthesis (what actually needs fixing).

## What does not change

Both skills are Jinja-gated on the profile; default renders are byte-for-byte what they were. The two agent files always install (with sentinel `model:` values when off) so `profile use` can flip a seat without a reinstall. No pipeline step, artifact path, or ship-gate check moved.

## Usage

```toml
[profile.fast-oracle]
extends = "full"
models = { fetcher = "cerebras/qwen-3.8-27b", synthesizer = "gemini/gemini-3.8-flash",
           decomposer = "openai/gpt-6-astra", corpus_critic = "openai/gpt-6-astra", chief_editor = "openai/gpt-6-astra" }
```

## Tests

- Roster test includes the two new agents.
- Defaults render no delegation text and sentinel model values.
- An overlay profile flips both skills into delegation mode and renders the model into agent frontmatter; other agents untouched.
- Chief editor prompt commits to all four findings files with the profile's per-lens caps and `{hpr_path}` substituted.
- ModelMap coverage test updated; full suite green locally (1,222 passed, 1 skipped), ruff clean.

https://claude.ai/code/session_01JyJz2SsnEZYWpyt55E7K3A


jordan-gibbs:
Closing; this was meant to stay local.

PR #124, opened 2026-09-12
fix: give the builtin provider a PDF lane, and say why a PDF was declined

### LINKED ISSUE #82: PDF extraction failures across independent mirrors trace to PyMuPDF parsing path
## Description
During a large batch run, multiple PDF extraction failures occurred across distinct mirrors for the same documents. Four independent hosting mirrors produced byte-identical `pymupdf` extraction errors. 

Because identical failure signatures occurred across completely separate remote hosts, this behavior indicates a parsing defect or unhandled edge case within the local PyMuPDF extraction pipeline rather than upstream host blocking, rate-limiting, or transport gating. Resolving this parser failure would eliminate downstream manual remediation waves (e.g., 117 affected papers in a recent run).

## Environment Details
- **Repository:** `jordan-gibbs/hyperresearch`
- **Component:** Local PDF extraction / PyMuPDF pipeline
- **Engine:** PyMuPDF (`fitz`)

## Observed Behavior
- Multiple independent mirror endpoints for identical documents fail during extraction.
- The resulting error logs/stack traces from PyMuPDF are byte-identical across all four mirror sources.
- The pipeline treats these as general extraction/retrieval failures rather than handling the underlying parser exception.

## Expected Behavior
- PDFs retrieved from valid mirrors should be correctly parsed by the local extraction worker.
- If a document triggers a known PyMuPDF syntax/structural edge case (e.g., specific stream encodings, xref repair, or encrypted headers), the pipeline should either recover gracefully, attempt a fallback extractor, or log a deterministic parser error rather than failing the run.

## Steps to Reproduce
1. Ingest a target document URL configured with multiple fallback mirror sources.
2. Trigger the automated retrieval and extraction workflow.
3. Observe extraction failure on the primary mirror.
4. Observe the exact same PyMuPDF error reproduced on the remaining 3 mirrors.

## Suggested Next Steps / Investigation
- Inspect the specific `pymupdf` / `fitz` exception block in the document processing worker.
- Verify whether the issue stems from malformed PDF xref tables, stream decoding, or version incompatibilities with specific PyMuPDF builds.
- Consider implementing an extraction fallback or pre-flight validation step for files encountering this parsing state.

jordan-gibbs:
## What this changes

The PDF lane (`_is_pdf_url` / `_fetch_pdf`) moves out of the crawl4ai provider into `web/pdf.py`, and the builtin provider uses it too. When the lane declines a URL and the HTML fallback turns out to be junk, `hpr fetch -j` now reports the lane's reason (`PDF lane: HTTP 403`, `no extractable text layer`, ...) instead of only the generic junk verdict. The pymupdf handle is closed on the exception path.

Fixes #82

## Why

A vault on `provider = "builtin"` (the default until `hpr install` switches it) had no PDF handling at all. A direct `.pdf` link was decoded as HTML text and rejected as "Binary PDF garbage in content", the same message for every mirror of the same document, which is the shape reported in #82. An arXiv `/abs/` link saved the 871-word abstract page instead of the paper, which is what @earldodd found while gathering diagnostics.

The lane's failure reasons went to `logging` with no handler configured anywhere in the package, so the operator only ever saw the junk message. The reason now travels on the result's metadata to the CLI.

## How it was verified

- `pytest tests/`: 1233 passed, 1 skipped (the POSIX signal test on Windows).
- `ruff check src/ tests/`: clean.
- New `tests/test_web/test_builtin_pdf_lane.py` (5 tests, offline through a stubbed `safe_get`): direct `.pdf` URL extracted; arXiv `/abs/` fetches the `/pdf/` URL and never touches the abstract page; lane failure falls back to HTML with the reason attached; PDF detected from bytes behind a `download?id=` URL; CLI junk verdict names the lane reason. All five fail on main.
- `tests/test_web/test_pdf_diagnostics.py` retargeted at the shared module, plus tests for the reason registry, the handle-close on iteration failure, and the crawl4ai aliases.
- Live, fresh builtin vault: `hpr fetch https://arxiv.org/abs/2005.14165` saves 38,147 words with `raw_file` set; on main it saves 871 words of abstract.

---

**Scope check**

- [x] No edits to user-facing output strings, docstrings, or test fixtures unrelated to this fix (the junk message gains the `PDF lane:` suffix only when a lane reason exists)
- [x] No formatter or automated cleanup pass applied

https://claude.ai/code/session_01JyJz2SsnEZYWpyt55E7K3A


earldodd:
Thank you!

PR #126, opened 2026-09-13
fix: note ids are fixed points of the slugifier; link validator rejects shell fragments and traversal

fduple:
## What this changes

`write_note` slugifies `note_id` at entry, so a note's filename always equals its frontmatter id and cannot contain a path separator. `is_valid_wiki_link_target` also rejects shell fragments and `.`/`..` path segments, and `repair` reports the stub targets it skips instead of swallowing the failure.

## Why

`write_note` slugified only the frontmatter id while building the filename from the raw `note_id`. `repair` and `graph stub` pass broken-link text verbatim, and that text is arbitrary fetched content:

- Bash test syntax such as `[[ -n "$kernel" ]]` parses as a wiki-link, and the stub loop writes literal files like `research/temp/-n "$kernel".md` (40 of these landed in one real vault from one Arch wiki page).
- A note body containing `[[../../outside]]` makes `hyperresearch repair` write `outside.md` at the vault root; one more `../` puts it outside the vault. The write is bounded (`.md` is always appended, the collision loop never overwrites, the body is a fixed stub), but it is still a path chosen by untrusted web content.

This is the same traversal class #125 closed for the run tag and #114 closed for `claims ingest --tag`, reached through note ids. It extends the content guards #114 added to `is_valid_wiki_link_target` (placeholders, unbalanced brackets), which do not catch these shapes. Braces stay with #114's placeholder check, which already rejects `${VAR}`. Refs shaped like real ids still pass by design.

## How it was verified

- 6 new tests in `tests/test_core/test_note.py` and `tests/test_core/test_patterns.py`. With the change reverted, the 4 guard tests fail on assertions: shell-fragment rejection, traversal rejection, filename/frontmatter agreement, and stays-in-vault (which actually writes `outside.md` outside the vault). The other 2 pin that clean ids and legitimate title-shaped refs (`UEFI/Secure Boot`, `C. elegans`, `yahoo! finance`) are unchanged.
- End to end on current main: `repair` on a note containing both refs above wrote the junk file and `outside.md`; with the fix it writes neither, and a second `repair` over ordinary refs creates 0 new stubs, so it stays idempotent.
- `pytest tests/`: 1275 passed on 3.11 and on 3.13. On 3.12, 1273 passed and the 2 tests in `tests/test_mcp/test_update_note.py` were skipped, because that local venv has `mcp` 2.x. `ruff check src/ tests/` is clean.

Happy to split the validator hardening from the `write_note` change if you prefer them separately; they close the same hole from two ends.

---


jordan-gibbs:
Merging. I reproduced the `../../outside` write on main, so thanks for finding that one, it's the same class as the run-tag hole I just closed in #125 and I didn't think to look at `write_note`.

I'll mention one side effect in the changelog: the validator also refuses title and alias links like `[[E=mc2]]` or `[[Ke$ha]]`, which used to resolve by title (`core/sync.py:381-397`) and will drop out of backlinks on the next sync. Slug ids can't hit it, so I'm happy taking that trade for now. If someone's vault actually leans on titles like that we can narrow it.

PR #127, opened 2026-09-13
security: classify IPv4-mapped IPv6 by its embedded IPv4; reject host:port allowlist entries

fduple:
## What this changes

The SSRF gate from #53 now classifies an IPv4-mapped IPv6 address (`::ffff:a.b.c.d`) by its embedded IPv4, and `allow_private_hosts` refuses a `host:port` entry loudly instead of accepting one that can never match.

## Why

- **Mapped CGNAT passes the gate on Python 3.11 and 3.13.** `_is_blocked_ip` relied on the `ipaddress` `is_*` properties to see through the mapping, and interpreters disagree on what those report. On 3.11.15 and 3.13.14, `check_url("http://[::ffff:100.64.0.1]/")` returns `None`: the carrier-grade NAT address is allowed and the OS connects to `100.64.0.1`. On 3.12.3 the opposite happens: every mapped address reports `is_reserved`, so a public destination like `[::ffff:8.8.8.8]` is refused.
- **Port-carrying allowlist entries silently never match.** `"192.168.1.20:8443"` and `"mirror.internal:8443"` were accepted as hostname entries, but URL hostnames never carry the port, so the user's mirror is refused and the refusal looks like the gate working as designed. A bracketed `"[::1]"` entry (written the way a URL writes it) had the same fate.

The fix unmaps first (`ip.ipv4_mapped`) and classifies the embedded IPv4, which gives the same verdict on every interpreter. A `host:port` entry now raises `SafeHTTPError` naming the entry, as a malformed CIDR already does, and brackets are stripped before an entry is parsed as an address. The comment next to `_6TO4_NET` claiming mapped addresses were already covered is removed.

## How it was verified

- 7 new tests in `tests/test_web/test_safe_http.py`: mapped private, metadata and CGNAT forms refused; mapped public allowed; both `host:port` shapes raise; a bracketed `[::1]` entry admits the literal.
- With the `safe_http.py` change reverted, 4 tests fail on each interpreter, and which guard fails depends on the interpreter:
  - 3.11 and 3.13: `test_check_url_rejects_ipv4_mapped_private_forms[http://[::ffff:100.64.0.1]/]` (the bypass);
  - 3.12: `test_check_url_allows_ipv4_mapped_public` (the over-refusal);
  - all three: both `test_allowlist_entry_with_port_is_a_loud_error` cases and `test_allowlist_bracketed_ipv6_entry_admits_the_literal`.
- `pytest tests/`: 1276 passed on 3.11 and on 3.13. On 3.12, 1274 passed and the 2 tests in `tests/test_mcp/test_update_note.py` were skipped, because that local venv has `mcp` 2.x. `ruff check src/ tests/` is clean.
- Only the three patch releases named above were measured. The fix does not depend on which behaviour an interpreter has.

---


jordan-gibbs:
Thanks for chasing this one. I checked it on 3.11 through 3.14, and mapped CGNAT (`::ffff:100.64.0.1`) really does get through on 3.12+ on main because the newer `is_*` properties see through the mapping. I didn't know that. Merging, and I'll add the changelog line on my side, including that a `host:port` allowlist entry now fails every fetch loudly instead of just the one host.

There's a sibling I missed back in 0.11.0: `check_oa_url` at `core/oa.py:134` still leans on `is_global`, and on 3.11 that calls `::ffff:100.64.0.1` global. I'll open an issue for folding it into `check_url`.

PR #128, opened 2026-09-13
fix: note mv resolves the destination into the synced tree, and refuses outside-tree and overwrite

fduple:
## What this changes

`hpr note mv` resolves its destination before touching anything. A bare name lands in `research/notes/` with `.md` appended; a destination outside `research/notes/`, `research/index/` or `research/temp/` is refused with `OUTSIDE_SYNCED_TREE`; an existing destination is refused with `DESTINATION_EXISTS`.

## Why

`note mv` joined its destination onto the vault root verbatim and renamed the file:

- **It overwrote notes.** In a fresh vault with `alpha-note` and `beta-note`, `hpr note mv beta-note research/notes/alpha-note.md` reports success; `alpha-note.md` now holds Beta's content, Alpha's content is gone, and the index row for `alpha-note` points at Beta's file.
- **A bare name dropped the note out of the index.** `hpr note mv beta-note beta-renamed` (what an agent naturally passes) moved the file to the vault root without `.md`, where sync never looks, and reported success.
- **Any outside-tree path was accepted**, including `../x` and absolute paths.

The containment check reuses the `_under` helper #114 added to `core/claims.py` for the `claims ingest --tag` traversal rather than adding a second one; #125's run-tag check is a slug pattern and does not apply to a path. JSON and console output now report the resolved path. The help text and docstring are corrected because they are about this behaviour: the old docstring promised mv updated all references, which nothing did (the id lives in frontmatter, so wiki-links keep resolving and nothing needs rewriting).

## How it was verified

- `tests/test_cli/test_note_ops.py`: `test_note_mv` now targets `research/notes/moved/` (its old target is outside the synced tree and is refused). New: a bare name lands indexed at `research/notes/<name>.md`; an outside-tree destination is refused with the source untouched and still indexed; moving onto an existing note is refused for both the explicit path and the bare name, with the target's content intact.
- With the `cli/note.py` change reverted, all 4 new cases fail on assertions; the other 8 tests in the file pass either way.
- Also checked by hand, each giving a clean error or the right path: `""`, `.`, `../x`, `/etc/passwd`, `research`, `research/notes`, `research/runs/t/x.md`, `research/foo.md`, an absolute path inside the vault, `research/index/i`.
- `pytest tests/`: 1273 passed on 3.11 and on 3.13. On 3.12, 1271 passed and the 2 tests in `tests/test_mcp/test_update_note.py` were skipped, because that local venv has `mcp` 2.x. `ruff check src/ tests/` is clean.

---


jordan-gibbs:
I pushed a commit on top of this instead of sending it back to you, since the fixes were small and I wanted it in with the rest of your batch.

`research/index/` came out of the allowed destinations. `IndexGenerator.build_all()` deletes every `.md` in there before regenerating (`indexgen/generator.py:17-19`), and `repair` calls it, so a note moved in was gone after the next repair. main had the same hole, I just hadn't noticed it until the new error message listed index as a place you could go.

The other one is Windows/macOS: a case-only rename (`beta-note` to `Beta-Note.md`) hit `DESTINATION_EXISTS` because `exists()` matches the note's own file on a case-insensitive filesystem. It checks `samefile` now, and `rel_new` resolves the parent only, since `resolve()` on the file hands back the old spelling. `.MD` also stopped picking up a second `.md`. Tests for all three are in `test_note_ops.py`.

Merging, thanks for this one.

PR #129, opened 2026-09-13
fix: builtin provider flattens fetched code to prose

fduple:
## What this changes

The builtin web provider keeps code as code: `<pre>` becomes a fenced block and inline `<code>` a backtick span, in both extraction branches (bs4 and the stdlib fallback). `BODY_DOI_RE` in `core/scholar.py` is adjusted so a DOI inside a code span is still found without a trailing backtick.

## Why

`BuiltinProvider._extract()` emitted bare `get_text()` output, so every code block reached the vault as plain lines. Beyond readability, the note parser strips fenced and inline code before extracting `[[wiki-links]]`, and bash test syntax is lexically a wiki-link, so a flattened shell script turns each conditional into a broken link that `graph stub` / `repair` then turn into junk notes. One Arch wiki page produced 40 stub notes named after shell fragments in a real vault, such as `! -f "$KERNELDESTINATION".md`.

Minimal reproduction on main:

```python
BuiltinProvider()._extract("<pre>[[ -n \"$kernel\" ]] || exit 0</pre>")[1]
# -> '[[ -n "$kernel" ]] || exit 0'   (no fence; one wiki-link)
```

Details: each fence is sized per block to outgrow any backtick run inside the code, so content cannot break out; inline spans double their backticks when needed; the stdlib branch exempts code from its whitespace collapse so indentation survives. `fetch()` is untouched: the SSRF gate and `max_html_bytes` cap in `safe_get()` still apply before any parsing, both PDF detections still run first, and the PDF lane's decline reason is still attached to the HTML result.

The DOI change is needed because of this fix: with code rendered as a span, `<code>DOI: 10.1000/xyz</code>` would give open-access recovery the DOI ``10.1000/xyz` `` (trailing backtick), and `DOI: <code>10.1000/xyz</code>` no DOI at all.

## How it was verified

- `tests/test_web/test_builtin_fences.py`: one contract class run over both extraction branches (fenced `<pre>`, backticked inline code, zero wiki-links after the note parser's exact strip sequence, indentation, fence lengthening, unchanged title/noise handling), plus two `fetch()`-level tests with `safe_get` stubbed, one of them an arXiv `/abs/` page reached after the PDF lane gets HTTP 403, which must keep both the fences and the "HTTP 403" decline reason.
- `tests/test_core/test_source_ranking.py`: a DOI-in-code-span case.
- With `builtin.py` and `scholar.py` reverted, 12 tests fail on assertions: 11 of the 14 builtin tests and the DOI test. The 3 that pass unpatched are guards for title/noise handling and bs4 indentation.
- `pytest tests/`: 1284 passed on 3.11 and on 3.13. On 3.12, 1282 passed and the 2 tests in `tests/test_mcp/test_update_note.py` were skipped, because that local venv has `mcp` 2.x. `ruff check src/ tests/` is clean.

---


jordan-gibbs:
I pushed one commit on top before merging. bs4's `replace_with()` and `decompose()` look the element up in its parent's child list, so a page with a lot of sibling `<code>` tags went quadratic: 50k siblings took 52.6s to extract, against 0.74s on main. Pages can be up to 10 MB, so a hostile one could pin a fetch for a long time. Clearing and refilling each element in place gives the same output and brings that page back to about 1.5s including the parse. Your 40 tests pass unchanged.

The changelog entry will mention one side effect. `strip_markdown` drops code before indexing, so code on builtin-fetched pages now leaves FTS, embeddings and the quote-integrity comparison. That's how crawl4ai pages already behaved, but a report quoting a sentence with an inline identifier can now fail quote integrity.

The fence can still be escaped downstream, and that part's on me. `CODE_BLOCK_RE` in `core/patterns.py` stops at the first triple-backtick run whatever the opening fence length, so code with an odd number of them still leaks links. It already leaked before this PR. I'll open an issue for it.

Merging, thanks. The Arch wiki stub-note thing had been bugging me.

PR #130, opened 2026-09-13
fix: the builtin provider raises CertVerificationError on a bad TLS certificate

fduple:
## What this changes

The builtin provider's HTML lane raises `CertVerificationError` when TLS certificate verification fails, as the PDF lane and the crawl4ai provider already do. Any other connection error propagates unchanged.

## Why

`BuiltinProvider._get` let the raw `httpx.ConnectError` from `safe_get` escape, so a certificate refusal surfaced as an opaque ssl string, indistinguishable by type from a refused connection or a DNS failure:

```python
BuiltinProvider(FetchSettings(allow_private_hosts=("127.0.0.1",))).fetch("https://127.0.0.1:<port>/page")
# httpx.ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate
```

**This is not a security bug**, and I checked rather than assumed. I traced a builtin certificate failure end to end through `hpr fetch` and `hpr fetch-batch` against a real self-signed server that counts connections and completed handshakes, with the open-access rescue forced to run and to hand back the refused host as its candidate. The raw error does reach the `except Exception` wrappers and the rescue, but nothing retries without verification: the server saw two connections per command and zero completed handshakes. Planting `verify=False` on the rescue's fetch raises the handshake count to one, so the check does detect an unverified retry. What was lost is only the distinct error type and message. The CLI has no certificate-specific error code, so `hpr fetch -j` still reports `FETCH_ERROR`, now with a message naming the certificate failure.

The fix re-raises an exception whose cause chain holds `ssl.SSLCertVerificationError` (the same `_is_cert_error` check the PDF lane uses) as `CertVerificationError`, chained to the original. Verification itself does not change. Every caller of the builtin provider catches `Exception`, so the new type is handled wherever `ConnectError` was.

## How it was verified

- 5 tests in the new `tests/test_web/test_builtin_cert_refusal.py`: a faked cert-error chain raises `CertVerificationError` after exactly one verified attempt; a non-certificate `ConnectError` propagates untranslated; a real loopback self-signed server gives `CertVerificationError` with one connection and no completed handshake; `hpr fetch -j` and `hpr fetch-batch -j` against that server with the rescue armed show the certificate message, two connections, zero handshakes.
- With the `builtin.py` change reverted, 4 of the 5 fail on behaviour; the pass-through test passes either way.
- The server fixture generates its certificate with `cryptography` (installed through Crawl4AI's pyOpenSSL dependency) and skips if it is missing.
- `pytest tests/`: 1274 passed on 3.11, 3.12 and 3.13. `ruff check src/ tests/` is clean.

---


jordan-gibbs:
Merging this. I ran it against the badssl hosts (expired, self-signed, wrong host, untrusted root) and they all come back as `CertVerificationError` now, with nothing retrying unverified.

One correction I'll carry into the changelog instead of the description: crawl4ai only raises this on its PDF lane. Its HTML lane runs with `ignore_https_errors=True` (`crawl4ai_provider.py:258`), which honestly I hadn't looked at until today and probably deserves its own issue.

PR #131, opened 2026-09-13
perf: vectorize semantic_search's cosine scan with numpy (#59)

### LINKED ISSUE #59: Proposal: pluggable semantic index backend for larger vaults
## Context

Semantic search currently stores float32 vectors in SQLite and scans them in Python in `src/hyperresearch/core/embed.py`. That is a good zero-configuration default and fits the existing design where Markdown is canonical and SQLite is a rebuildable cache.

Would the project welcome a small semantic-index boundary so larger vaults can opt into an external vector index without changing the default behavior?

## Possible shape

- Keep the current SQLite BLOB plus brute-force cosine implementation as the default.
- Define a narrow backend contract for vector sync, upsert, delete, and search.
- Keep embedding provider selection, note metadata, FTS, and reciprocal-rank fusion in the existing project layers.
- Load external backends only through optional dependencies and explicit configuration.
- A first optional backend could be Milvus, using local Milvus Lite where supported or a configurable remote URI and token.

## Questions

1. Does a pluggable semantic-index boundary fit the intended architecture?
2. Should external indexes remain derived and fully rebuildable from the Markdown vault and SQLite metadata?
3. If this direction is welcome, would you prefer the backend contract to live in core or in a separately installable integration package?

This issue is only to confirm the architecture and contribution route before any implementation.

SID-6921:
## Summary

Closes #59 by doing the part the maintainer asked for first ("vectorize first, then revisit... that's the PR I'd merge tomorrow"):

- `cosine()` scored one candidate at a time in pure Python (three generator passes per call). `semantic_search` now loads every stored vector with `np.frombuffer` and scores the whole set in a single matmul. `cosine()` itself is untouched and still covered by its own test — this only changes how `semantic_search` uses the stored vectors.
- A note still holding a vector from a since-changed embeddings model/dimension (mid-migration, before the next `embed sync` catches it up) now scores `0`, the same way `cosine()` did on a length mismatch, instead of joining the matrix and blowing up the reshape.
- `numpy` is now a declared direct dependency. It was already an undeclared transitive dependency via `Crawl4AI`, so this adds no new install footprint.
- Corrected the module docstring's and the roadmap doc's "fine to ~50k notes" claim — per the maintainer's own measurements in #59, that was off by about an order of magnitude, having been written as if the vectorized path already existed.
- Added a CHANGELOG entry under `[Unreleased]`.

## Test plan

- [x] New regression test compares `semantic_search`'s ranking/scores against an independently-computed, per-row `cosine()` loop over the same stored vectors (not just re-running `semantic_search` against itself).
- [x] New regression test confirms a dimension-mismatched vector scores `0` instead of crashing.
- [x] `python -m pytest tests/ -v` — full suite passes (270+ tests).
- [x] `ruff check src/ tests/` — clean.
- [x] Standalone benchmark (not part of the test suite): 5,000 random 512-dim vectors, naive Python loop vs. the new matmul path — same top-5 ranking and scores within `1e-4`, ~25x faster.


jordan-gibbs:
Scores match the old loop to within float noise, same ordering, and I measured 15–19x end to end. #119 just landed and touches `semantic_search` (it joins `notes` so filters apply before the limit), so this needs a rebase onto it. While you're in there:

- the docstring's "hundreds of thousands of notes" overstates it; at 300k × 1536 it's seconds and about 5 GB peak. Something like "sub-second to ~50k notes" is accurate.
- `np.sqrt(np.einsum("ij,ij->i", mat, mat))` instead of `np.linalg.norm` roughly halves the time and the peak memory at 50k.
- a test with a zero vector would cover the `where=` branch.

PR #132, opened 2026-09-16
Fix/118 non blocking crawl4ai fetch

### LINKED ISSUE #118: fetch_url/fetch_many block the event loop during a fetch (async def + anyio.to_thread.run_sync)
`fetch_url` and `fetch_many` in `web/crawl4ai_provider.py` wrap their async work with a blocking `_run_coro` helper that joins the thread from the event-loop thread, so the MCP server serves nothing for the duration of a fetch (raised by @jordan-gibbs in #86, follow-up requested there).

Making `fetch_url` `async def` and awaiting `anyio.to_thread.run_sync` directly would keep the server responsive during a fetch. It touches the provider's fetch signature more than #86's scope, which is why it stayed out of that PR.

Not something I've built yet, opening this so it doesn't get lost.

akshaypal912:
## Summary

Fixes #118 by removing the blocking coroutine/thread-join pattern from the Crawl4AI MCP fetch path.

### Changes

* Convert Crawl4AI `fetch_url` to an async entry point.
* Convert Crawl4AI `fetch_many` to async and update its sync CLI boundary.
* Remove the blocking `_run_coro` usage from the Crawl4AI provider.
* Use `anyio.to_thread.run_sync` for blocking Crawl4AI/PDF work where needed.
* Make the MCP `fetch_url` tool async and use `fetch_and_save_async`.
* Preserve synchronous `fetch()` for CLI usage and explicitly reject calls from an already-running event loop.
* Add `anyio>=4.0` to the `mcp` extra because the async fetch/save path imports AnyIO directly.

### Tests

Added/updated tests covering:

* Awaiting `fetch_url` on a running event loop.
* Event-loop responsiveness during a Crawl4AI fetch.
* Explicit failure when synchronous `fetch()` is called from a running event loop.
* MCP `fetch_url` awaiting the async save path.
* MCP event-loop responsiveness.
* Packaging dependency coverage for the `mcp` extra.

### Validation

* `python -m pytest tests/test_web/ -v` → **147 passed**
* `python -m pytest tests/ -q` → **1230 passed, 2 skipped**
* `ruff check` → **passed**

### Compatibility

* Async callers should use `await fetch_url(...)` / `await fetch_many(...)`.
* Synchronous CLI callers continue using `fetch()` / the batch CLI.
* `fetch()` now raises an explicit `RuntimeError` if called from an already-running event loop; async callers should use `fetch_url()` instead.

The Crawl4AI provider no longer blocks the MCP event loop by synchronously joining a worker thread during fetch operations.


jordan-gibbs:
This conflicts with main now, and #145 (the #137 fix) just changed the same crawl4ai provider code. Can you rebase onto main? I'll review then.

akshaypal912:
ok jordan-gibbs , i will do it ....

